PR #266, opened 2014-07-16
WIP: Adding ClearDisplay

zaxtax:
Trying to add primitive so display may be cleared from Haskell.


gibiansky:
This does not seem to be getting much activity and I have not checked whether it works, so closing this. Will open a new issue for this functionality though.

PR #317, opened 2014-12-28
[WIP] make custom.js work with 3.0

Carreau:
Hi, 

These are the modification I had to do for custom.js to work on 3.0, 
need a little more work to have them also work on 2.3 without changes, 
but just in case you want to try. 

You should "just" need the following also: 
```
$ cat .ipython/kernels/haskell/kernel.json
{"display_name":"IHaskell","argv":["/Users/bussonniermatthias/.cabal/bin/IHaskell" ,"kernel", "{connection_file}"]}
```

And it should work with IPython 3.0.

Side note, conceal extension is quite cool.

got a kernel crash with `\.<tab>`
```
[I 14:55:55.516 NotebookApp] Saving file at /Untitled.ipynb
IHaskell: src/IHaskell/IPython/Message/Parser.hs:(125,3)-(129,59): Irrefutable pattern failed for pattern Data.Aeson.Types.Internal.Success parsed
```

You can easily reproduce programatically with `IPython.notebook.kernel.complete('\.',2, function(data){console.log(data)})`

Should I open a separate bug ?

### DIFF
--- a/profile/static/custom/custom.js
+++ b/profile/static/custom/custom.js
@@ -1,4 +1,10 @@
-$([IPython.events]).on('notebook_loaded.Notebook', function(){
+console.warn('haskell custom js loading')
+require(['require', 'codemirror/lib/codemirror','codemirror/addon/mode/loadmode', 'base/js/namespace','base/js/events', 'base/js/utils'],function(require, CodeMirror, CodemirrorLoadmode, IPtyhon, events, utils){
+
+
+
+events.on('notebook_loaded.Notebook', function(){
+    console.warn('haskell custom js set metadata');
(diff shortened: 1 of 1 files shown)


gibiansky:
Ack! This somehow slipped through the cracks and I didn't notice it until now. Please open a separate bug for the crashing! I'll take a look at this soonish, I would like to make sure IHaskell works with IPython 3.0 shortly after IPython 3.0 is released.

Carreau:
> Ack! This somehow slipped through the cracks and I didn't notice it until now. Please open a separate bug for the crashing!

No need to apologize, last weeks were pretty crazy and next will be to. 
I still need to polish this PR.

PR #328, opened 2015-01-03
fix test

cblp:
This makes test compilable, but test still fails with
```
Exception: Failed to load interface for `IHaskell.Display'
```

### DIFF
--- a/src/Hspec.hs
+++ b/src/Hspec.hs
@@ -59,7 +59,7 @@ eval string = do
 
   getTemporaryDirectory >>= setCurrentDirectory
   let state = defaultKernelState { getLintStatus = LintOff }
-  interpret "" False $ Eval.evaluate state string publish
+  interpret libdir False $ Eval.evaluate state string publish
   out <- readIORef outputAccum
   pagerOut <- readIORef pagerAccum
   return (reverse out, unlines $ reverse pagerOut)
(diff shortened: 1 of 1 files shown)


gibiansky:
The first argument to `interpret` is the GHC library directory, obtainable via `GHC.Paths.libdir` (if you import `GHC.Paths`. Can you change `""` to `GHC.Paths.libdir` and add `import qualified GHC.Paths` in the toplevel module? 

Thank you for finding this mistake – was some recent changes that broke the test suite, which has been failing for a while now (needs to be cleaned up and fixed...)

cblp:
Oh, it can be fixed even simpler.

gibiansky:
thanks!

cblp:
But the test still fails:
```
1) Code Evaluation evaluates expressions
uncaught exception: ErrorCall (Expecting value!)

2) Code Evaluation evaluates multiline expressions
uncaught exception: ErrorCall (Expecting value!)

3) Code Evaluation evaluates function declarations silently
uncaught exception: ErrorCall (Expecting value!)

4) Code Evaluation evaluates data declarations
uncaught exception: ErrorCall (Expecting value!)

5) Code Evaluation evaluates do blocks in expressions
uncaught exception: ErrorCall (Expecting value!)

6) Code Evaluation is silent for imports
uncaught exception: ErrorCall (Expecting value!)

7) Code Evaluation evaluates directives
uncaught exception: ErrorCall (Expecting value!)
```

gibiansky:
Yeah – that is an older issue. I am not sure why it fails there... I originally opened https://github.com/gibiansky/IHaskell/issues/272 for that, but haven't had the time to deal with it.

Issue #236, opened 2014-04-30
Allow inline TemplateHaskell


gibiansky:
See previous discussion on https://github.com/gibiansky/IHaskell/issues/183.

Implement once GHC 7.8 is being used, as then we can use `pprint`? Or perhaps `zeroth`.

cje:
Has anybody managed to get DecQ splices working under GHC 7.8.3 and IHaskell?   It seems like with that, acowley/Frames would work.

gibiansky:
No, this hasn't been worked on at all. I've talked to @houshuang about some ideas for how to implement it so I think there is a pretty clear path towards the implementation, but no effort has been put into it yet. (If you're interested in taking a stab at it, let me know, and we can discuss implementation strategies. Otherwise, hopefully I'll get to it eventually.)

gibiansky:
It's definitely something I'm very interested in getting working though – without it, you can't really use things like lenses or various special deriving things in IHaskell...

Issue #183, opened 2014-01-31
Is template haskell supposed to work ?


miguel-negrao:
From lens help file:

```
:ext TemplateHaskell
data Foo a = Foo { _bar :: Int, _baz :: Int, _quux :: a }
makeLenses ''Foo
```

this make the kernel busy...

PR #330, opened 2015-01-05
Permit use of inline template haskell

cje:
Addresses Github issues: https://github.com/gibiansky/IHaskell/issues/236  , https://github.com/gibiansky/IHaskell/issues/183

To verify functionality, paste the following into a code block in
IHaskell.

```
:ext TemplateHaskell
import Language.Haskell.TH
import Control.Lens
data Foo a = Foo { _bar :: Int, _baz :: Int, _quux :: a }
makeLenses ''Foo
:ty baz
```

### DIFF
--- a/src/IHaskell/Eval/Evaluate.hs
+++ b/src/IHaskell/Eval/Evaluate.hs
@@ -699,7 +699,6 @@ evalCommand output (Expression expr) state = do
   -- typeclass instance, this will throw an exception and thus `attempt` will
   -- return False, and we just resort to plaintext.
   let displayExpr = printf "(IHaskell.Display.display (%s))" expr :: String
-  write displayExpr 
   canRunDisplay <- attempt $ exprType displayExpr
 
   -- Check if this is a widget.
--- a/src/IHaskell/Eval/Evaluate.hs
+++ b/src/IHaskell/Eval/Evaluate.hs
@@ -79,7 +79,7 @@ data ErrorOccurred = Success | Failure deriving (Show, Eq)
 
 -- | Enable debugging output
 debug :: Bool
-debug = False  
+debug = False    
 
 -- | Set GHC's verbosity for debugging
 ghcVerbosity :: Maybe Int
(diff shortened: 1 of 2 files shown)


src/IHaskell/Eval/Evaluate.hs
```diff
@@ -704,8 +705,20 @@ evalCommand output (Expression expr) state = do
   let widgetExpr = printf "(IHaskell.Display.Widget (%s))" expr :: String
   isWidget <- attempt $ exprType widgetExpr
 
+  -- Check if this is a declaration  
+ -- let declExpr = printf "((id :: Q [Dec] -> Q [Dec]) (%s))" expr::String
```
gibiansky: Is this just a comment leftover from debugging?

src/IHaskell/Eval/Evaluate.hs
```diff
@@ -716,20 +729,31 @@ evalCommand output (Expression expr) state = do
...
-
-    let out = evalResult evalOut
-        showErr = isShowError out
-
-    -- If evaluation failed, return the failure.  If it was successful, we
-    -- may be able to use the IHaskellDisplay typeclass.
-    return $ if not showErr || useShowErrors state
-            then evalOut
-            else postprocessShowError evalOut
+    if isDeclaration
```
gibiansky: We'll want to switch around the order of these if blocks – I'll explain in the PR discussion briefly.

gibiansky:
To elaborate on my previous point. We would like the `DecsQ` check to happen *before* the `IHaskellDisplay` check. If we do this, then we can have an instance

```haskell
instance IHaskellDisplay DecsQ where
   display = ...
```
which generates code using TH and then displays it. 

IHaskell by default would just run the generated code, but if you used `display` directly you could instead see it, e.g. `display (makeLenses ''MyData)` or something. In order to do that, DecsQ would have an IHaskellDisplay instance, so if the IHaskellDisplay instance is checked first, then the DecsQ one would never be used. However, if the DecsQ is checked first, you can still force IHaskellDisplay to be used via `display`; so checking for DecsQ first would yield more flexibility.

gibiansky:
This doesn't actually completely work. If you don't `import Language.Haskell.TH`, this should also work, right? Right now, it says that there's no `Show` instance for `DecQ`...

gibiansky:
That however is easily fixed by pre-importing TH qualified; should be fixed in `master`.

PR #332, opened 2015-01-07
Add IHaskell.Display.IHaskellDisplay instance for DecsQ

cje:


gibiansky:
I'm getting the error
```
Template Haskell error: Can't do `reify' in the IO monad
```
in the console and the error
```
user error (Template Haskell error)
```
in the notebook. 

What's going on?

gibiansky:
The issue is with `TH.runQ decl` – that is the line that throws the error. How should this be fixed?

gibiansky:
It looks like this is a well known error... Is there any way around this?  Perhaps writing Haskell code to a file, then using something like `zeroth`, then reading from that file? (That approach used to be used with `hlint` until the new API came about.)

cje:
That's odd - that fix worked fine for me testing.  What's the expression
you're evaluating?  I'll have a look shortly.

gibiansky:
Anything which calls `reify`, I think. I used `makeLenses ''Foo`. (Simpler things, like `[| \x -> x + x |]`, work, I believe.)

gibiansky:
Actually, not with the example given above, since that's `ExpQ`. `ExpQ` could also have an instance that is very similar, right?

What example did you use for testing?

PR #335, opened 2015-01-09
Add automation for easily defining kernels

david-christiansen:
The module `IHaskell.IPython.EasyKernel` provides automation for writing
simple IPython kernels. In particular, it provides a record type that
defines configurations and a function that interprets a configuration as
an action in some monad that can do IO.

The configuration consists primarily of functions that implement the
various features of a kernel, such as running code, looking up
documentation, and performing completion. An example for a simple
language that nevertheless has side effects, global state, and timing
effects is included in the examples directory in a file called Calc.hs.

Presently, there is no automation for creating the profile in the
.ipython directory. One should follow the IPython instructions for this.

### DIFF
--- a/ipython-kernel/examples/Calc.hs
+++ b/ipython-kernel/examples/Calc.hs
@@ -1,0 +1,236 @@
+{-# LANGUAGE FlexibleContexts, FlexibleInstances, OverloadedStrings, MultiWayIf #-}
+module Main where
+
+import Control.Applicative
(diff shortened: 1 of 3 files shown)


gibiansky:
Looks good. Clean design, good code. Merging.

david-christiansen:
Thanks! If I have a few extra hours tomorrow I'll get going on IIdris :-)

PR #339, opened 2015-01-13
Support installing profile

david-christiansen:
Now, EasyKernel supports installing a .tar file containing it's ipython
profile, which includes information about how to launch the executable.


gibiansky:
Looks good.

That said, note that all of this is likely to change soon in the coming months with Jupyter / IPython 3.0. 

Finally, now that this exists, what are the limitations of `EasyKernel`? What does IHaskell have that `EasyKernel` does not implement?

david-christiansen:
I suppose I'll have to keep it up to date for 3.0 - the docs make it look quite different, though better.

EasyKernel doesn't support installing ipython or anything like that. Also, it might be hairy to implement the fancy display methods that IHaskell has, or to do other more reflective things, though this could perhaps be improved by making the display and completion functions into actions in the configuration's monad.

EasyKernel is designed to make it quick to make something reasonable, and some flexibility is sacrificed. If we can keep the easy API and make it usable as a basis for IHaskell, much the better, however.

gibiansky:
Alright, got it. 

The most likely scenario in 3.0 is that the installation bit of IHaskell is going to be removed entirely, as will the profiles. Instead, profiles are likely to be replaced with kernel profiles. Effectively, instead of IHaskell managing IPython, it will be swapped, and IHaskell will just install its kernel into the IPython/Jupyter directories. We'll see how it goes though – architecture discussions are still ongoing and things may change a little before 3.0...

david-christiansen:
If you keep me posted, then this EasyKernel module should be fairly straightforward to update for that style.

PR #341, opened 2015-01-25
Experimental Display instance for Frames 

cje:
Note that as Frames is not yet on hackage, the cabal file cannot be completed, and this should not be merged just yet.


houshuang:
What's the status of this, do you have a demo? I was trying to get it to work with IHaskell (very interested in getting this working!), I checked out your branch and cabal installed the ihaskell-frames directory. However, when I do :browse IHaskell.Display.Frames in ghci, it shows me nothing, and it doesn't seem to work when I load some of acowley's demos in IH. 

```haskell
No instance for (Show (Producer Income IO ())) arising from a use of ‘print’
In a stmt of an interactive GHCi command: print it
```

cje:
It's working, but there are a few moving parts which need to be sorted out to make the build process automatic.  

Can you please try the following and let me know how you go?: 
 - Install Frames (looks like this is already done given the message above) 
 - Build IHaskell using the shell  "./build.sh display" (rather than calling cabal in ihaskell-frames) 
 - Retry the example above? 

In order for the process to be automatic, we need to publish Frames on hackage, but that's not a good idea until the API is more stable, as changes could break other packages. [@acowley]

gibiansky:
@cje 

Do you mind closing this PR and opening a new one? PRs are now automatically checked by Travis to make sure they build and all tests pass, and Travis isn't enabled for PRs that existed before Travis was set up. 

Also, note that `./build.sh` has changed behaviour -- if you want to do a 100% full install, you can do `./build.sh all`, while `./build.sh display` doesn't build `ghc-parser` or `ipython-kernel` (which are probably irrelevant in your case, but figured I'd let you know that the interface has changed a bit; just run `./build.sh` to see new interface command list.)

cje:
Andrew - No problem.

Issue #316, opened 2014-12-27
shelly $ which "ipython" -> Empty File path.


Carreau:
Please excuse my limitted knowlege of Haskell, 

I finally achieve to build IHaskell on one of my machine to try to help migrate to IPython 3.0.

Starting any of IHaskell command greats me with : 
```
ihaskell:
Ran commands:
which ipython

Exception: Empty filepath
```
Or alike emty filepath  for `which IHaskell`

which I tracked down to Shelly:
```
Prelude Shelly> shelly $ which "ipython"
*** Exception:
Ran commands:
which ipython

Exception: Empty filepath
Prelude Shelly> shelly $ which "/usr/local/Cellar/python3/3.4.2_1/Frameworks/Python.framework/Versions/3.4/bin/ipython"
Just (FilePath "/usr/local/Cellar/python3/3.4.2_1/Frameworks/Python.framework/Versions/3.4/bin/ipython")
```

I also found out that in `ipython_config.py` the `kernel_command = [exe, ...]`  did not had the `exe` replaced, but I fixed that by hand. 

I suppose this is just a stupid issue with $PATH as in new bash shell if I reverting to IPython 2.3.1 I can start things correctly, but I can't figured out why Shelly stil refuses to locate IPython and/or IHaskell sometimes. 

Anyway, the 4 lines of Haskell I wrote represent the majority of Haskell lines I have written in my life and i'll probably figured out how to fixed that by hardcoding the path in the source for now, 
but maybe you have a better idea. 

Also in IPython.hs ~ line 180 you seem to try to handle the case of shelly returning nothing: "Did not detect IHaskell-installed or system IPython." but did not catch this shelly error (or whatever equivalent)

Maybe that would be a good idea ? 

Anyway documenting my first encounter with a IHaskell that compile (yeah).

And also wow, feel super weird to revert to IPython 2.3.1.

Issue #271, opened 2014-08-01
Failed running after successful installation on Mac OS X 10.9


neolee:
Hi gibiansky,

IHaskell is great idea! But unfortunately I've never made it work on my Mac OS X. Here are some issue I've met.

1 . Followed the guide in the project README: 
 - Install gcc, haskell-platform, zeromq and python via Homebrew (surely some of those were already there for a while);
 - Install ipython through pip (and all its dependencies);
 - Update cabal and install cairo, happy, cpphs using cabal;
 - Finally install IHaskell using cabal. All done without errors.

2 . Running `IHaskell console` gives error:
```
IHaskell:
Ran commands:
which ipython

Exception: Empty filepath
```

3 . Running `IHaskell console -i /usr/local/bin/ipython` gives error:
```
[ZMQTerminalIPythonApp] ERROR | Exception while loading config file /Users/neo/.ipython/profile_haskell/ipython_config.py
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/site-packages/IPython/config/application.py", line 514, in _load_config_files
    config = loader.load_config()
  File "/usr/local/lib/python2.7/site-packages/IPython/config/loader.py", line 425, in load_config
    self._read_file_as_dict()
  File "/usr/local/lib/python2.7/site-packages/IPython/config/loader.py", line 478, in _read_file_as_dict
    py3compat.execfile(conf_filename, namespace)
  File "/usr/local/lib/python2.7/site-packages/IPython/utils/py3compat.py", line 224, in execfile
    builtin_mod.execfile(filename, *where)
  File "/Users/neo/.ipython/profile_haskell/ipython_config.py", line 4, in <module>
    c.KernelManager.kernel_cmd = [exe, 'kernel', '{connection_file}']
NameError: name 'exe' is not defined
[IPKernelApp] ERROR | Exception while loading config file /Users/neo/.ipython/profile_haskell/ipython_config.py
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/site-packages/IPython/config/application.py", line 514, in _load_config_files
    config = loader.load_config()
  File "/usr/local/lib/python2.7/site-packages/IPython/config/loader.py", line 425, in load_config
    self._read_file_as_dict()
  File "/usr/local/lib/python2.7/site-packages/IPython/config/loader.py", line 478, in _read_file_as_dict
    py3compat.execfile(conf_filename, namespace)
[...]

PR #348, opened 2015-01-28
info on setting path to ihaskell in ipython config file

stared:
There is one extra step for installation (at least on OS X), which took me some time.

It was already solved by https://github.com/gibiansky/IHaskell/issues/271 (as an intermediate step) and in https://github.com/gibiansky/IHaskell/issues/316.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,11 +141,6 @@ If you already have ipython installed you can use the --ipython flag.
 
 On Linux make sure you also have `python-dev` (or equivalent) installed, which is needed to install `pyzmq`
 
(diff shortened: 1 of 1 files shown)


gibiansky:
This is a bug, actually, not a documentation issue. That line should be inserted by IHaskell itself upon installation/first run...

It may however nonetheless be useful to note that as a "possible problem/fix"...

stared:
I moved it to Troubleshooting (aka "Gotchas") section. In any case, do you know what is the root cause / how can it be fixed?

gibiansky:
This is going away very soon in the `jupyter` branch, where the profile will be replaced with a `kernelspec`, so closing this.

PR #353, opened 2015-02-06
GHC 7.10 support

bgamari:
Introduce parser from GHC commit a993712d16691829c3bf6d1ea552112abb2d4f9f.

I have yet to test `ihaskell-charts`, `ihaskell-diagrams`, `ihaskell-hatex`, and `ihaskell-rlangqq`.


bgamari:
Unfortunately this isn't sufficient to get things working. It seems there are some parser issues as any attempt to pass input to IHaskell results in the error `Module must have a name.`

bgamari:
It seems that the trouble here is that all input is parsed to a Template Haskell splice in a module. For instance, the input `print "hi"` is parsed to `$(print "hi")`. I'm not sure why this is just now happening with GHC 7.10. Perhaps I missed something while importing the parser?

gibiansky:
So, there's a chance that we can avoid making a new parser for GHC 7.10. 

The issue is probably that the old parsers are modified, I *think*, to forbid modules that don't start with a module header `module ... where ....`.

The parser module for 7.10 should include *most* of the things necessary for parsing Haskell code without resorting to `ghc-parser` style madness; it should have a few added exports that are used by `ghc-parser`. That said that still leaves the module problem which may require a bit of unusual trickery.

(My goal was to stop cloning parsers in 7.10, so I added the exports to GHC, but I haven't tried updating ghc-parser to use the updated exports when dealing with 7.10)

bgamari:
I'm very glad to hear that the `ghc-parser` approach is on its way out. I've pushed a commit fixing the `ghc-parser`7.10 port. I'm not yet sure whether it will be otherwise possible to reproduce `namedModule` without duplicating the parser. That being said, it seems like this would be something that could be sent upstream.

gibiansky:
Indeed. Sadly, it looks like I didn't forsee this. It's possible it'd still be possible to slip this into the GHC 7.10, but I sadly don't have the time for this right now...

That said, it may also be possible to check via other means, e.g. check for the substring "module" in a non-string non-quasiquoted context, or parse it as a module and then check if it has a module head.

gibiansky:
I think it should be possible to do this without copying the GHC 7.10 parser. Specifically, we can just try parsing the block of code as a module and then checking whether the `hsmodName` field of `HsModule` is `Nothing`; if it *is*, then the module is unnamed, and we artificially fail the parsing. See [HsModule documentation here](https://www.haskell.org/platform/doc/2013.2.0.0/ghc-api/HsSyn.html).

This is cheating a little bit but may work. 

Also, note that the `ghc-parser` package is not going anywhere – it will still provide a nice wrapper for the GHC API. The (Parser module)[http://hackage.haskell.org/package/ghc-parser-0.1.3.0/docs/Language-Haskell-GHC-Parser.html] will still be useful, just not the HappyParser module.

bgamari:
Every time I commit on `master` I eventually come to regret it.

PR #357, opened 2015-02-17
Adding TH to Cabal's test-suite requires

houshuang:
Seems that TH needs to be listed in the Cabal file for the test-suite to run.


gibiansky:
Ah, thanks for pointing this out! TH actually shouldn't be necessary at all. Removing that import and the dependency. Should be fixed now.

PR #359, opened 2015-02-17
Ignoring > in front of code lines

houshuang:
Ignores all > at the beginning of lines, which makes it easier to copy and paste code from literate Haskell files (like this: https://github.com/VinylRecords/Vinyl/blob/master/tests/Intro.lhs). 

Don't think > can ever occur in Haskell code at the beginning of a line, so it should be safe.

### DIFF
--- a/src/IHaskell/Eval/Evaluate.hs
+++ b/src/IHaskell/Eval/Evaluate.hs
@@ -230,9 +230,10 @@ data EvalOut = EvalOut {
   }
 
 cleanString :: String -> String
-cleanString x = strip $ if allBrackets then clean else x
+cleanString x =  if allBrackets then clean else str
   where
-    l = lines x
+    str = strip x
+    l = lines str
     allBrackets = all (fAny [isPrefixOf ">", null]) l
     fAny fs x = any ($x) fs
     clean = unlines $ map removeBracket l
--- a/src/IHaskell/Eval/Evaluate.hs
+++ b/src/IHaskell/Eval/Evaluate.hs
@@ -237,7 +237,7 @@ cleanString x =  if allBrackets then clean else str
     allBrackets = all (fAny [isPrefixOf ">", null]) l
     fAny fs x = any ($x) fs
     clean = unlines $ map removeBracket l
-    removeBracket (x:xs) = xs
+    removeBracket ('>':xs) = xs
     removeBracket [] = []
     -- should never happen:
     removeBracket other = error $ "Expected bracket as first char, but got string: " ++ other 
--- a/src/IHaskell/Eval/Evaluate.hs
+++ b/src/IHaskell/Eval/Evaluate.hs
@@ -233,11 +233,13 @@ cleanString :: String -> String
 cleanString x = strip $ if allBrackets then clean else x
   where
     l = lines x
-    allBrackets = all (fAny [isInfixOf ">", null]) l
+    allBrackets = all (fAny [isPrefixOf ">", null]) l
     fAny fs x = any ($x) fs
     clean = unlines $ map removeBracket l
     removeBracket (x:xs) = xs
     removeBracket [] = []
+    -- should never happen:
+    removeBracket other = error $ "Expected bracket as first char, but got string: " ++ other 
 
 -- | Evaluate some IPython input code.
 evaluate :: KernelState                  -- ^ The kernel state.
--- a/src/IHaskell/Eval/Evaluate.hs
+++ b/src/IHaskell/Eval/Evaluate.hs
@@ -229,13 +229,23 @@ data EvalOut = EvalOut {
     evalComms :: [CommInfo]
   }
 
+cleanString :: String -> String
+cleanString x = strip $ if allBrackets then clean else x
+  where
+    l = lines x
+    allBrackets = all (fAny [isInfixOf ">", null]) l
+    fAny fs x = any ($x) fs
+    clean = unlines $ map removeBracket l
+    removeBracket (x:xs) = xs
+    removeBracket [] = []
+
 -- | Evaluate some IPython input code.
 evaluate :: KernelState                  -- ^ The kernel state.
          -> String                       -- ^ Haskell code or other interpreter commands.
          -> (EvaluationResult -> IO ())   -- ^ Function used to publish data outputs.
          -> Interpreter KernelState
 evaluate kernelState code output = do
-  cmds <- parseString (strip code)
+  cmds <- parseString (cleanString code)
   let execCount = getExecutionCounter kernelState
 
   when (getLintStatus kernelState /= LintOff) $ liftIO $ do


gibiansky:
Is there a good reason to pull in a pretty hefty dependency on a regex engine for this? Could this instead be implemented as follows:

```haskell
cleanString :: String -> String
cleanString = strip . unlines . map deBird . lines
  where deBird ('>':x) = x
        deBird x = x
```
In general I would like to avoid pulling in new dependencies, unless they are somewhat justified – this case seems like it might be fairly easy to do?

gibiansky:
Will be happy to merge this if the new dependency is removed – I think this is a good feature to have.

One other discussion question: perhaps it makes sense to do this only if all the lines start with `<`? Not sure where though. Specifically, this would avoid the issue of quasiquoting... What happens if someone enters:

```haskell
let x = [myRandomQuasiquoter|
> Hello!
> Goodbye!
This is a special language
|]
```
This is valid Haskell and could do something special. But, on the other hand, I can't think of a reason why *every* line would have a `>` as the first character, in which case it would be safe.

houshuang:
Hey,

yes it always feels a bit dirty in Haskell to do regexp, however in my
defence, the Data.String.Utils library, which you already depend on,
depends on this library and uses it

{- | Splits a string around whitespace.  Empty elements in the
resultlist are automatically removed. -}splitWs :: String ->
[String]splitWs = filter (\x -> x /= []) . splitRegex (mkRegex "[
\t\n\r\v\f]+")

:) Presumably this means that IHaskell won't have to pull in a new
dependency, since it already depends (recursively) on Regex.Compat.

However, happy to do it the other way as well. Since we're not talking much
text, I guess efficiency isn't very important either way (I'm not sure how
much the difference would be, if any). Makes sense to check if all lines
have a quote in front of them. I'll take a new stab at it.

(It would be neat to have an LHS importer that turns text into text blocks
and > into code blocks, but that's a separate issue :)

On Mon, Feb 16, 2015 at 10:26 PM, Andrew Gibiansky <notifications@github.com
> wrote:

> Will be happy to merge this if the new dependency is removed – I think
> this is a good feature to have.
>
> One other discussion question: perhaps it makes sense to do this only if
> all the lines start with <? Not sure where though. Specifically, this
> would avoid the issue of quasiquoting... What happens if someone enters:
>
> let x = [myRandomQuasiquoter|
> > Hello!
> > Goodbye!This is a special language|]
>
> This is valid Haskell and could do something special. But, on the other
> hand, I can't think of a reason why *every* line would have a > as the
> first character, in which case it would be safe.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/gibiansky/IHaskell/pull/359#issuecomment-74612005>.
>



-- 
http://reganmian.net/blog -- Random Stuff that Matters

gibiansky:
I believe that Adam Vogt wrote a LHS importer. See `IHaskell.Convert.LhsToIpynb`. Sadly this is likely to be invalidated upon the switch to IPython 3...

And yes, `Data.List.Split` does already incur that dependency, but it still helps to minimize direct dependencies – because then we have less lower and upper bounds to worry about. Plus, it seems like a fairly small change :)

As for efficiency, yeah, we haven't worried about that too much yet. There's definitely some speed issues – there's a momentary but noticeable and annoying delay between evaluation and output. I don't know why, though, and haven't put effort into profiling it. But I doubt it's the text handling...

src/IHaskell/Eval/Evaluate.hs
```diff
@@ -229,13 +229,23 @@ data EvalOut = EvalOut {
     evalComms :: [CommInfo]
   }
 
+cleanString :: String -> String
+cleanString x = strip $ if allBrackets then clean else x
+  where
+    l = lines x
+    allBrackets = all (fAny [isInfixOf ">", null]) l
```
gibiansky: Do you mean `isPrefixOf`?

```haskell
isInfixOf ">" "a > b" == True
```

src/IHaskell/Eval/Evaluate.hs
```diff
@@ -229,13 +229,23 @@ data EvalOut = EvalOut {
...
+    fAny fs x = any ($x) fs
+    clean = unlines $ map removeBracket l
+    removeBracket (x:xs) = xs
```
gibiansky: Just for safety purposes, why not change this to
```haskell
removeBracket ('>':xs) = xs
removeBracket [] = []
removeBracket other = error $ "Expected bracket as first char, but got string: " ++ other
```

houshuang:
You are right about infix/prefix, changed it. When it comes to removeBracket, it is only run once it has been verified that every line starts with either > or is [], but if you want to add it as a check, that's fine. Safety first :) I still feel like the regexp solution I had on three lines (including checking that all lines begun with >) was cleaner, but happy to get this in anyway, very useful functionality.

src/IHaskell/Eval/Evaluate.hs
```diff
@@ -229,13 +229,26 @@ data EvalOut = EvalOut {
...
 
+cleanString :: String -> String
+cleanString x =  if allBrackets then clean else str
+  where
+    str = strip x
+    l = lines str
+    allBrackets = all (fAny [isPrefixOf ">", null]) l
+    fAny fs x = any ($x) fs
+    clean = unlines $ map removeBracket l
+    removeBracket (x:xs) = xs
```
gibiansky: This should be 

```haskell
removeBracket ('>':xs) = xs
```
instead of

```haskell
removeBracket (x:xs) = xs
```
Otherwise the `other` case is never reached.

gibiansky:
Almost perfect :) Fix that very last issue (`x` -> `'>'`) and it's ready to merge.

Just to explain my reasoning for the two changes.

1. Suppose we do pull in a regex package. Since we need to ensure it works with IHaskell, we give it upper and lower version bounds. `regex < 2 && regex >= 1`. Then, `regex` updates to version 3. Alright, that's fine. Everything should still build! But now `split` starts using something from `regex-3.1`, so it regures `regex >= 3`. Now, unless `split` nicely behaved and properly updates version numbers, `ihaskell` doesn't compile anymore, and does so silently, without us having changed anything. Rather than deal with this in a large number of packages, it's easier to just avoid new packages unless they incur significant benefit. (That said, I wonder if the dependency list of `IHaskell` could be trimmed down some...) If that's not convincing, call it personal preference :)

2. The same sort of logic goes for `removeBracket`. If there is some change to the condition checking (we want to optimize the text handling, perhaps), it's better if `removeBracket` fails early and fails loudly, rather than perhaps letting the error propagate later and yield confusing things when we try to debug it.

Thanks!

houshuang:
Thanks for all the explanations. Yes, Haskell package management is crazy. I tried nuking my package directory, and starting from scratch with Stackage. After installing only IHaskell and Frames, `ghc-pkg list | wc` shows 253 packages... That all have to be compatible.

PR #360, opened 2015-02-17
Adding show instances to CommInfo and EvaluationResult

houshuang:
Makes interactive debugging of the evaluate command easier.


gibiansky:
thanks

PR #368, opened 2015-02-23
Adding empty case to set

houshuang:
Better than having kernel panic on typing ":set" without a flag... Not sure if this should have a help text or an error text.

### DIFF
--- a/demo/IHaskell.ipynb
+++ b/demo/IHaskell.ipynb
@@ -318,6 +318,7 @@
      "cell_type": "markdown",
      "metadata": {},
      "source": [
+      "You can use option :no-pager to force the information to appear inline instead, which is useful for including in demonstrations that you intend to share with others.\n",
       "We can now write slightly more complicated scripts."
      ]
     },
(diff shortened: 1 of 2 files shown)


gibiansky:
Good catch. I'm happy with what it currently says there.

aavogt:
ghci interprets that as "what options are set":
```
> :set
options currently set: none.
base language is: Haskell2010
with the following modifiers:
  -XNoDatatypeContexts
  -XNondecreasingIndentation
GHCi-specific dynamic flag settings:
  -fbreak-on-exception
other dynamic, non-language, flag settings:
  -fimplicit-import-qualified
warning settings:
  -fwarn-monomorphism-restriction
```
I think we should follow ghci's behavior if there's no reason to do things differently

gibiansky:
Hmm, agreed. I will still keep this merged as an error message is better than a crash, but the ghci version is probably the desired behaviour.

houshuang:
@bgamari seems to have already implemented this. https://github.com/bgamari/IHaskell/commit/9abaee86cdb573dcc7ff40cf6169e25b7dd91b2a

bgamari:
@houshuang indeed I have. I've split out this change from the rest of the GHC 7.10 changes and opened pull request #375.

PR #376, opened 2015-02-24
GHC 7.10 support

bgamari:
This supersedes #353.

Introduce parser from GHC commit a993712d16691829c3bf6d1ea552112abb2d4f9f.

I have yet to test ihaskell-charts, ihaskell-diagrams, ihaskell-hatex, and ihaskell-rlangqq.


gibiansky:
Does that mean that the other `display` packages have been tested? If so, I will go ahead and merge this. I do not have GHC 7.10 yet so cannot verify that everything works as desired, so would just test on 7.8 and would then test on 7.10 when it is released.

bgamari:
Well, I just started looking at the parser issue that we discussed in #353.  I would like to avoid copying the GHC parser if possible however I'm not sure I see how it would be possible provide the partial parsers.

bgamari:
For what it's worth the other `display` packages have indeed been tested.

gibiansky:
If I remember correctly, the partial parsers are actually never used, and so should actually be removed. Specifically, I believe partial parses are just treated as errors (see `Parser.hs:165` or so).

The main issue with not copying the GHC parser will be detecting if a module header exists, and only treating it as a module if a header does exist.

bgamari:
@gibiansky ahh, lovely. In that case I'll just try dealing with the module header as you described in #353.

bgamari:
@gibiansky I've opened up #379 which implements the approach you described.

bgamari:
It looks like #379 will work out. Closing this.

Issue #380, opened 2015-02-24
Compilation failure with FlagSpec


FranklinChen:
Current `master` head does not compile on GHC 7.8.4:

```
src/IHaskell/Eval/Util.hs:81:17:
    Not in scope: type constructor or class ‘FlagSpec’

src/IHaskell/Eval/Util.hs:84:17:
    Not in scope: type constructor or class ‘FlagSpec’
```

gibiansky:
@bgamari This is due to the `:set` PR. How does one fix this (where is `FlagSpec` from)?

gibiansky:
Fixed this by removing the type signatures. Not optimal but works.

bgamari:
@gibiansky This is rather odd; `FlagSpec`should be exported from `DynFlags` on 7.8.

bgamari:
Ahh, never mind; it's defined but not exported. Removing the type signatures may in fact be the least bad option.

PR #381, opened 2015-02-25
Eval.Util: Define FlagSpec type synonym

bgamari:
Apparently this wasn't exported pre-7.10.

Properly fixes breakage reported in #380 due to #375.

### DIFF
--- a/src/IHaskell/Eval/Util.hs
+++ b/src/IHaskell/Eval/Util.hs
@@ -78,7 +78,12 @@ extensionFlag ext =
     flagMatchesNo ext (name, _, _) = ext == "No" ++ name
 
 #if !MIN_VERSION_ghc(7,10,0)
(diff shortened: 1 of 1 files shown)


bgamari:
Still waiting for the test builds to finish.

gibiansky:
This should not work; `TurnOnFlag` and `DynP` are not imported. (I tried the same thing when trying to fix this and ended up getting rid of the imports instead of tracking down all the types, sorry)

bgamari:
@gibiansky right, indeed the build failed. I have the type tracked down but just omitting them may be the least bad solution.

bgamari:
@gibiansky indeed, even the types that I tracked down aren't sufficient. Given that this is just to handle the older GHC releases, I say we just remove the type signatures for now.

gibiansky:
Sounds good to me. Leaving things as they are now.

PR #383, opened 2015-02-26
Adding pretty-show by default

houshuang:
screen shot 2015-02-26 at 10 10 09

Making Text.Show.Pretty the default Show method. (An alternative would be to make it optional, but then ':set -InteractiveShow' needs to be implemented, currently ghci doesn't pick it up when used in IHaskell, even if it gives no error)

### DIFF
--- a/ihaskell.cabal
+++ b/ihaskell.cabal
@@ -76,6 +76,7 @@ library
                        MissingH             >=1.2,
                        mtl                  >=2.1,
                        parsec               -any,
+                       pretty-show          -any,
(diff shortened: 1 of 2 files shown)


gibiansky:
Closing for the time being, as a) it can no longer be merged and b) I would like to use Travis CI for this PR, as it is now set up for testing, but it only checks new PRs (this one was made before I set up Travis).

That said I am happy to accept the variant of this contribution that we discussed if you do end up getting around to it!

PR #384, opened 2015-02-26
Upping boundary on Shelly, to make IH compatible with Stackage

houshuang:


gibiansky:
What is the standard for libraries in Stackage? Do they use `==` or `>=` for dependencies?

houshuang:
They don't care, what happens (according to my understanding) is that it
parses all the requirements of all the cabal files, and comes up with a
build-plan that satisfies all packages (otherwise the pull requests will
fail, and they won't be added). In that build plan, which can be downloaded
as a cabal.conf file, every single package is pinned to a specific version.
But the larger range of versions each individual package allows, the easier
it is to come up with the build plan. In this case, it has already
chosen shelly-1.6.1.1
as the one version that fulfills all existing packages' requirements, so to
add IHaskell, we have to have a range which makes that version acceptable
to IHaskell. But we don't need to pin the version - a new version might be
the "optimal version" for all packages in the future.

We've already dealt with the other one (mono-traversable), so with this
change, we should be compatible with Stackage.

(PS If you push to Hackage, could you make a tag on Github? Makes it easier
to diff later).

On Thu, Feb 26, 2015 at 5:39 PM, Andrew Gibiansky <notifications@github.com>
wrote:

> What is the standard for libraries in Stackage? Do they use == or >= for
> dependencies?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/gibiansky/IHaskell/pull/384#issuecomment-76290679>.
>



-- 
http://reganmian.net/blog -- Random Stuff that Matters

gibiansky:
Got it. Sounds good. How was mono-traversable dealt with?

houshuang:
We bumped it in cbcae02

On Thu, Feb 26, 2015 at 7:14 PM, Andrew Gibiansky <notifications@github.com>
wrote:

> Got it. Sounds good. How was mono-traversable dealt with?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/gibiansky/IHaskell/pull/384#issuecomment-76304188>.
>



-- 
http://reganmian.net/blog -- Random Stuff that Matters

PR #393, opened 2015-03-03
Removing unused imports, fixing other -fwarn issues

houshuang:
Seems to be a bunch of left-over imports and unused functions in Parser.hs. Still compiles cleanly after taking all this out. Not sure if this kind of pull-request is welcome? I think it makes it easier to navigate the source for others if all the imports are used, etc.


houshuang:
Also making some pattern-matches exhaustive.

gibiansky:
Yes, code cleanliness PRs are welcome. I would like IHaskell to compile with -Wall *except* for the name shadowing warnings but haven't gotten around to it.

PR #394, opened 2015-03-03
minimal change to get IHaskell kernel.js loaded at the right time

Carreau:
Not indented, I'll strip that down, but you should not need the first on notebook loaded I guess.

### DIFF
--- a/html/kernel.js
+++ b/html/kernel.js
@@ -1,12 +1,14 @@
-console.log('Kernel haskell kernel.js is loading.');
 
-require(['require',
+define(['require',
         'codemirror/lib/codemirror',
         'codemirror/addon/mode/loadmode',
         'base/js/namespace',
         'base/js/events',
         'base/js/utils'],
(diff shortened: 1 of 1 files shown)


gibiansky:
Thanks! This doesn't seem to change whether or not `app_initialized.NotebookApp` is called though. I also tried `notebook_loaded.Notebook`, but that doesn't seem to be called either...

gibiansky:
Do you know of any examples of kernel.js from other kernels? I looked around but could not find any

gibiansky:
Oh, I understand. Got it.

Carreau:
Yeah, I'm writing the docs for it right now.

Carreau:
Ok, strugling with git book, but seem to be online now : http://carreau.gitbooks.io/jupyter-book/content/kerneljs.html  does that make sens ?

gibiansky:
Got it, thanks. Makes sense!

PR #403, opened 2015-03-05
On MacOS, build.sh installs arithmoi without llvm

FranklinChen:
Add constraint on cabal install for MacOS to not use LLVM for arithmoi, addressing a known problem:

https://bitbucket.org/dafis/arithmoi/issue/15/fails-to-compile-with-ghc-784-due-to-fllvm

http://stackoverflow.com/questions/24796874/cant-install-diagrams-arithmoi-on-mac

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -57,7 +57,12 @@ done
 
 # Stick a "./" before everything.
 INSTALL_DIRS=`echo $INSTALLS | tr ' ' '\n' | sed 's#^#./#' | tr ' ' '\n'`
-cabal install -j $INSTALL_DIRS --force-reinstalls
+
+if [ `uname` = Darwin ]; then
+  cabal install --constraint "arithmoi -llvm" -j $INSTALL_DIRS --force-reinstalls
+else
+  cabal install -j $INSTALL_DIRS --force-reinstalls
+fi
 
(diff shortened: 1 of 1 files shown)


gibiansky:
Do you mean `-fllvm`? `-llvm` I think is enabled by default already, no? (I might be misreading things...)

FranklinChen:
`cabal install --constraint "arithmoi -llvm" ...` is correct.

`cabal install 'arithmoi==0.4.*' -f -llvm` for installing just `arithmoi` is also correct.

By the way, just a few seconds ago I finally completed a fully automated install of IHaskell on Mac OS on a fresh machine. It was quite involved (since my normal machines had a lot of state already from fiddling around with things) but I got it working and will verify it by wiping everything and running my script.

Right now I have the setup do various things before finally calling `./build.sh all` and `./build.sh display`. Should I just submit this wrapper script as a specific `mac-build-everything.sh` or something like that?

gibiansky:
Sounds good. Yes, wrapper script sounds good -- that said, make sure there is good documentation about what the script does (even if just as comments in the script), because as a user, I would be afraid of running such a script, given that it has to do a lot of things (install native libs, etc).

PR #417, opened 2015-03-07
Improve Mac, Linux installation scripts. Still assumes using build.sh.

FranklinChen:
Linux: rid using `python-pip` since it eventually caused an insidious problem: http://stackoverflow.com/questions/27341064/how-do-i-fix-importerror-cannot-import-name-incompleteread 

Mac: add installation of IPython.

### DIFF
--- a/linux-install.sh
+++ b/linux-install.sh
@@ -1,16 +1,38 @@
 #!/bin/sh
 
-# Installation for Linux (tested on Ubuntu 14.10)
+# Installation for Linux (tested on Ubuntu 14.10) from IHaskell repo directory.
+# TODO Split out setup for installation from Hackage released versions.
 
-sudo apt-get install python-dev
-sudo apt-get install python-pip
-sudo pip install -U 'ipython[all]'
+# Install IPython.
+# python-pip is out of date, causes problems, so we get the latest version.
+#sudo apt-get install python-pip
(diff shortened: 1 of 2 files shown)


gibiansky:
Looks good, but I don't think it's a good idea to install `ghc` and `cabal` from `hvr`'s apt repositories. In general I'm a little leery of adding apt repositories, and it seems like the goal of those is really for testing rather than for end users. I would rather require users to already have a proper version of GHC and cabal installed, and instead just check that they exist, rather than installing them. In particular, installing them makes this script less useful to people who *already* have them, which is many.

So I think this is good save those changes to `linux-install.sh`.

FranklinChen:
You're right, `ghc` and `cabal` should not be installed by part of the script, but should be installed already by the user. In the Mac script, I didn't install them, so I got inconsistent! Changing the Linux script now...

gibiansky:
Looks good. One last small change -- sorry for the hassle -- `build.sh` has changed behaviour such that `build.sh all` builds literally verything, including widgets. So the `build.sh` display at the end should be removed entirely. Once that's done I will go ahead and merge.

FranklinChen:
OK, I've removed the `build.sh display`.

gibiansky:
Hooray! The first PR that Travis approves of.

PR #421, opened 2015-03-07
Ugh, fixed a couple of things in Linux install.

FranklinChen:
Argh, this is what happens when something works and then I make some tweaks but don't run the resulting script as it stands.

### DIFF
--- a/linux-install.sh
+++ b/linux-install.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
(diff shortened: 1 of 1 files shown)


gibiansky:
Can you edit this to only commit `linux-install.sh` and not the `HappyParser.hs` changes?

gibiansky:
(Unless of course they are relevant?)

FranklinChen:
Ugh, I restored the files that get regenerated during builds.

PR #422, opened 2015-03-08
Fix IHaskell convert for ipynb files with nbformat 4.

juselius:
``IHaskell convert`` failed to to convert  .ipynb files with nbformat 4 to lhs. The back transformation produced v3 files, which could subsequently be converted to v4 upon loading into the notebook. These patches fix the issues, making ``convert`` operate on v4 files only.

### DIFF
--- a/src/IHaskell/Convert/LhsToIpynb.hs
+++ b/src/IHaskell/Convert/LhsToIpynb.hs
@@ -3,12 +3,12 @@
 module IHaskell.Convert.LhsToIpynb (lhsToIpynb) where
 
 import Control.Applicative ((<$>))
-import Data.Aeson ((.=), encode, object, Value(Array, Bool, Number, String))
+import Data.Aeson ((.=), encode, object, Value(Array, Bool, Number, String, Null))
 import qualified Data.ByteString.Lazy as L (writeFile)
 import Data.Char (isSpace)
 import Data.Monoid (Monoid(mempty))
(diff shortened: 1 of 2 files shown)


juselius:
I was a bit optimistic. lhs to ipynb does not quite work yet.

src/IHaskell/Convert/LhsToIpynb.hs
```diff
@@ -1,4 +1,5 @@
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE CPP #-}
```
gibiansky: What do we need CPP for?

gibiansky:
Cool. Let me know when lhs to ipynbV4 is ready to merge.

I would also love to have some tests of the functionality in `Hspec.hs`, as that is now actually checked with Travis CI, but I'm also happy to just file that as an issue once this is merged and add it later.

src/IHaskell/Convert/LhsToIpynb.hs
juselius: VERSION_ghc is defined as a cpp macro. I'll have a fully working version soon.

gibiansky: Ah, got it! I scanned through the source for `#endif` or other directives but saw nothing; didn't think of the cabal macros.

juselius:
Ok, now everything seems to work!

PR #427, opened 2015-03-10
Fix empty Bird style code lines.

juselius:
Empty Bird style code lines, i.e. lines containing only a single '>' were
treated like markdown. Only code lines with '> ' were treated like empty code
lines.


juselius:
A small note: This patch is not "perfect", since one could envision a markdown cell containing a line with a single '>' in it. The line would mistakenly be interpreted as a code cell. However, in a notebook chances are far greater that '>' means code than markdown. Implementing a more robust version would probably require both lookahead and transformation history. I think it's overkill, so until somebody files an issue, I'll leave it as it is.

gibiansky:
Sounds good. Thanks for the note. I personally don't use this feature, so am okay with leaving it as it is until this becomes an issue, if it ever does.

PR #438, opened 2015-03-15
Added `ihaskell install` into build.sh

slon1024:


gibiansky:
Good call. Will merge once travis verifies build.

gibiansky:
Hmm, looks like this doesn't quite work. The issue is that `ihaskell install` requires `ipython >3.0`, but `./build.sh ihaskell` should be able to exit with 0 exit status regardless. 

Do you mind fixing this so that `./build.sh` does not fail if `ipython` is not present? For example, by returning a 0 exit code even if `ihaskell install` fails.

gibiansky:
Ready to merge?

slon1024:
Yeap :smile:

PR #450, opened 2015-03-25
Add IHaskellDisplay instance for Figure ()

sumitsahrawat:
plot

Allows easy usage of [plot](https://hackage.haskell.org/package/plot) package inside ihaskell.


gibiansky:
Awesome! But it looks like it doesn't build on Travis, because it doesn't install blas or lapack. Could you update the Travis configuration to build properly? Edit .Travis.yml in the top level directory and add the necessary install commands to get it to build. Let me know if you have any trouble or questions.

gibiansky:
Also, could you update linuxinstall.sh and macosxinstall.sh as well with the proper install commands ?

sumitsahrawat:
@gibiansky This should work. I have updated `linux-install.sh` and `.travis.yml` config file, but not the `macos-install.sh` script. I don't have access to a mac, and travis wouldn't verify whether the script is correct, so I can't make changes to it with any reliability.

gibiansky:
Sounds good. I can deal with the `macos-install.sh` fixes; most of my development is on a Mac.

sumitsahrawat:
@gibiansky All the 'build' related issues have been resolved. The only trouble is with the formatting.
It seems hindent is leaving a erratic empty line at the end of file. It seems like I'll have to use hindent repository's head for formatting.

gibiansky:
Merged (via a `rebase`, since it couldn't automerge). Style checks out. (FYI the style is checked using `hindent` HEAD; the most recent changes to it aren't on Hackage yet). All looks good!

That said, I haven't used `plot`, but sometimes if you have polymorphic values like `Figure a`, they won't be able to display with an instance for `Figure ()`, because the `a` is ambiguous. So the display instances often provide a type-specified version of `id` which you have to apply to your value to get rid of the polymorphic type (instead of a type annotation). For example, in `ihaskell-diagrams`:

```haskell
-- Rendering hint.
diagram :: Diagram Cairo R2 -> Diagram Cairo R2
diagram = id
```
I don't know if you need one for `plot`, but just pointing out this potential difficulty.

Issue #458, opened 2015-03-26
linux-install.sh breaks gentoo systems and only works on ubuntu


triclops200:
Not all linux systems use apt-get. Also, easy-install is buggy on some linuxes, and using it on Gentoo is enough to effectively destroy users' systems. Please find a cross-platform way to fix this.

gibiansky:
Fair enough; sounds like we should rename it to `ubuntu-install.sh`. I'm fairly explicit about that [here in the README](https://github.com/gibiansky/IHaskell#ubuntu).

@FranklinChen Do you mind renaming it?

gibiansky:
This has been clarified in the naming and the README. I don't have a Gentoo system so can't provide an install script that will work, so closing this issue (and the recommended installation path is via the other routes). That said I am happy to accept PRs for more install scripts.

PR #459, opened 2015-03-26
Rename linux-install.sh to ubuntu-install.sh

FranklinChen:
To clarify that the script is meant for Ubuntu Linux. https://github.com/gibiansky/IHaskell/issues/458


gibiansky:
Thanks! Also just updated the docs for this in the README.

Issue #347, opened 2015-01-28
Type on mouseover


oliverbunting:
It would be really nice it type on evalulated cells could be shown on mouseover, similar to ide's such as atom.

houshuang:
+1

gibiansky:
Comments from IPython devs on this:

**Me**:
> Get some info about an identifier from the kernel when I hover my mouse over that identifier, and display it as a popup.
> Idea: Use CodeMirror API and IPython comms to communicate with kernel.

**Response:**
> I think CM.coorsChar is the relevant API.
>
> You might not need to use comms for this one, if `inspect_request` is the right action. If the output should be distinguished from inspection via other means, then a custom comm would be appropriate.

sumitsahrawat:
I studied the API provided by codemirror, and these are the interesting functions:

- `cm.getTokenTypeAt`: A function that does the job partially. i.e. it tells the type of a token in the syntax-tree sense (null for unstyled tokens, an array containing all applicable styles otherwise). (Good for a initial proof-of-concept)
- `cm.getTokenAt`: Retrieves information about the token in the current mode found before the given position. (includes the type, string, start-end positions, and the mode's parser state at end of this token).

Assuming that there is a way to extract the type of any token in a code fragment in IHaskell (using the GHC API), I am now thinking about building a proof-of-concept that just shows the type (AST sense type).

For displaying the type, I am thinking of using the `display/panel.js` addon from CodeMirror to show a thin panel at the bottom of the cell, which will automatically hide when inactive. The problem with it is that  it might require frontend support. Another possible way is to provide a widget similar to `parsec-widget`, but that will look clumsy.

Also, I did not get the idea about `inspect_request`. There is no such function provided by CodeMirror. Is it a part of the `IPython/Jupyter` messaging specification?

gibiansky:
@sumitsahrawat 

That sounds promising. For `inspect_request`, check out that portion of the IPython messaging spec [here](http://ipython.org/ipython-doc/dev/development/messaging.html#introspection).

I suggest as a starting point, just get it to show (as a tooltip) whatever symbol the mouse is hovering over at the time. One that works, then you can use javascript comms to talk to the backend.

sumitsahrawat:
The messaging spec seems to have changed. IHaskell implements `object_info_request` which has been  renamed to `inspect_request`. I'll rename some constructors to make it match the latest spec.

gibiansky:
[...]

PR #468, opened 2015-04-03
Prevent kernel death on introspection.

sumitsahrawat:
I have updated the code relating to introspection so that it now respects the messaging spec v5.0.
Still, there is no tooltip shown on the frontend, even though I was able to confirm (using `Debug.Trace`) that the message produced by `toJSON` (from `IHaskell.IPython.Message.Writer`) had proper structure.

I searched the ipython repository for keyword `tooltip`, and it seems like it is implemented in IPython but not in jupyter. We might have to implement custom tooltips for IHaskell.

This will be required for #347.

### DIFF
--- a/ipython-kernel/src/IHaskell/IPython/EasyKernel.hs
+++ b/ipython-kernel/src/IHaskell/IPython/EasyKernel.hs
@@ -90,7 +90,7 @@ data KernelConfig m output result =
          , completion :: T.Text -> T.Text -> Int -> Maybe ([T.Text], T.Text, T.Text)
          -- | Return the information or documentation for its argument. The returned tuple consists of the
          -- name, the documentation, and the type, respectively.
-         , objectInfo :: T.Text -> Maybe (T.Text, T.Text, T.Text)
+         , inspectInfo :: T.Text -> Maybe (T.Text, T.Text, T.Text)
          -- | Execute a cell. The arguments are the contents of the cell, an IO action that will clear the
          -- current intermediate output, and an IO action that will add a new item to the intermediate
          -- output. The result consists of the actual result, the status to be sent to IPython, and the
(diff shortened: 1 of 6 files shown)


sumitsahrawat:
## More details

For an example cell,

```haskell
sin pi
```

Tracing `toJSON` for `PublishDisplayData`, we get:

```json
{
    "data": {
        "text/plain": "1.2246467991473532e-16"
    },
    "metadata": {},
    "source": "haskell"
}
```

Tracing `toJSON` for `InspectReply`, with inspection on `pi` gives:

```json
{
    "status": "ok",
    "data": {
        "text/html": "<pre><b>Type:<b> forall a. Floating a => a</pre>",
        "text/plain": "forall a. Floating a => a"
    },
    "metadata": {}
}
```

According to the [spec](http://ipython.org/ipython-doc/dev/development/messaging.html#introspection), the `data` should be a mime-bundle similar to `display_data`, which it is as seen above.

sumitsahrawat:
@gibiansky  The ipython devs (@Carreau) on gitter told me that just sending a `inspect_reply` message should work. I have done that but I think I'm missing something. Can you take a look?

Also, introspection on the master branch causes the kernel to die. With this code, it doesn't. This seems to be a step in the right direction.

gibiansky:
Alright, will do!

Carreau:
To be more clear it should work on shift-tab, otherwise the handler won't know where to put the message.

Can be changed though.

sumitsahrawat:
You mean to say that if the `inspect_reply` is sent without first receiving a `inspect_request` (sent by a shift-tab), the tooltip would not show, right?

Currently, my code doesn't work even on shift-tab, and I have been unable to find a reason for it.

Thanks for the help.

gibiansky:
What's the state of this PR?  Specifically:

- What exactly does it enable?
- Is it ready to merge?
- If not, what still needs to be done to get it merge-ready?

sumitsahrawat:
> What exactly does it enable?

Currently, this receives the input (whole cell text + cursor position), finds the word at that position (no proper lexing), and prints a lot of info, including the proper types in the console if `kernelDebug = True`. I also implemented proper replies, but the tooltip doesn't show for some unknown reason.

> Is it ready to merge?

Not yet.

> If not, what still needs to be done to get it merge-ready?

1. Proper lexing needs to be done, most probably using the GHC API.
2. The tooltip has to appear.

I tried tracing the incoming-outgoing json messages and nothing looks fishy. I also changed `replyTo` to send a fixed reply, ignoring the contents of the request and yet it didn't work.

sumitsahrawat:
I've edited the PR so that it prevents kernel death on introspection. The tooltip still does **not** show, but this can be merged as it stands.

gibiansky:
Perfect. I will investigate and merge today

sumitsahrawat:
Having a strange issue on travis. `./build.sh ihaskell` works for me (GHC 7.10.1). Travis shows an error in a very unexpected place.

gibiansky:
The build failures were unrelated I believe to this PR. Some of them may have been fixed in [this one](https://github.com/gibiansky/IHaskell/pull/488) which I just merged (I did not realize it was ready to merge). There may also have been some changes to `classy-prelude` which are breaking things, but I got things to build fine; I'll investigate travis issues separately from this PR.

Can I go ahead and merge this, then?

sumitsahrawat:
You can go ahead and merge this. I don't have anything else I can add.
You could also try restarting the build, now that #488 has been merged.

gibiansky:
@sumitsahrawat It turned out that you needed a `found` key to be true in order for the tooltip to display. This was completely undocumented (although about a week ago there was an ipython PR to document this, but the documentation has not yet appeared in the messaging spec online in HTML form). Once you added a `found` key everything displayed and your PR worked.

PR #471, opened 2015-04-07
set global CM Mode

Carreau:
I think this should work for you. I'm trying to get a backport so that `set('null')` is not needed.
And the loop on all cell should not be needed once you drop 2.x support.

You mode is simple enough though, I would also send it upstream to codemirror.

### DIFF
--- a/html/kernel.js
+++ b/html/kernel.js
@@ -10,7 +10,7 @@ define(['require',
             var onload = function(){
                 console.log('Kernel haskell kernel.js is loading.');
 
-                // add here logic that shoudl be run once per **page load**
+                // add here logic that should be run once per **page load**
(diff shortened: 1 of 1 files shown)


Carreau:
and cross link with https://github.com/ipython/ipython/pull/8271

gibiansky:
Does CodeMirror accept arbitrary modes? Is the `ipython` mode in the codemirror upstream?

Carreau:
> Does CodeMirror accept arbitrary modes?

Most of changes I proposed where accepted. 
I pushed some config options to python mode (make `?, !` identifiers as option), it was accepted. 

We plan to push the rest of the mode upstream too. Somme things to figure out to dynamically swithch `%%magics` highlighting in magics though.

PR #473, opened 2015-04-14
Allow using pre-installed ipython

sumitsahrawat:
Pinging @FranklinChen for comments.

### DIFF
--- a/ubuntu-install.sh
+++ b/ubuntu-install.sh
@@ -3,33 +3,35 @@
 # Installation for Linux (tested on Ubuntu 14.10) from IHaskell repo directory.
 # TODO Split out setup for installation from Hackage released versions.
 
-ghc --version >& /dev/null
-if [ $? ]; then
-    true
-else
+if [ -z "$(which ghc)" ]; then
     echo "Please install ghc."
+    exit 1
 fi
 
-cabal --version >& /dev/null
-if [ $? ]; then
-    true
-else
+if [ -z "$(which cabal)" ]; then
     echo "Please install Cabal."
+    exit 1
 fi
 
-# Install IPython.
-# python-pip is out of date, causes problems, so we get the latest version
-# using easy_install instead.
-#sudo apt-get install python-pip
-sudo apt-get install -y python-dev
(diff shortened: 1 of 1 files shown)


gibiansky:
@sumitsahrawat Is this for a hypothetical `Dockerfile` based on `ipython/ipython` or `jupyter/minimal` or something?

sumitsahrawat:
I am in the process of editing the `Dockerfile` in `jupyter/docker-demo-images`. With this, installing IHaskell will be straightforward, just clone the repo, and execute `docker-common.sh`.

gibiansky:
Then we'll have IHaskell on try.jupyter.org, I think. Awesome.

gibiansky:
Would it be possible to merge this into one script that uses a conditional?

For example, check whether `ipython --version` returns 3.0 or above; if it doesn't, install it, and if it does, do nothing. 

That way, we can just use `ubuntu-install.sh` for Docker as well.

Something along these lines:
```bash
if [ `ipython --version | sed 's#\.# #g' | cut -f 1 -d ' '` -ge 3 ]; then
   # don't install ipython
else
  # install ipython
fi
```

sumitsahrawat:
That looks like a better option, I'll update the PR.

sumitsahrawat:
Another thing I noticed is that the docker containers use `ipython3` as opposed to `ipython`. Would IHaskell be able to work with that?

gibiansky:
IHaskell should not care at all.

FranklinChen:
Makes sense to check for IPython and not forcibly install it.

gibiansky:
@sumitsahrawat Is this tested and ready-to-merge? (Note that I do not have Ubuntu on hand so cannot test this easily.)

sumitsahrawat:
I'll squash these commits, I'm experimenting with this script in a virtual machine.

sumitsahrawat:
This works. Tested with and without pre-installed ipython. Also tested with a lower version of ipython (v2.0) pre-installed.

gibiansky:
Great. Will merge shortly.

Issue #476, opened 2015-04-18
There is only 7.6 support in hackage


ztq56:
In https://hackage.haskell.org/package/ghc-parser the package depends on `ghc` (>=7.6 && <7.11) but only `src-7.6` folder in the cabal source package.
I install failed on my arch linux ghc 7.10.1, cabal 1.22.2.0, and It shows:

> Resolving dependencies...
Downloading ghc-parser-0.1.6.0...
Configuring ghc-parser-0.1.6.0...
Building ghc-parser-0.1.6.0...
Failed to install ghc-parser-0.1.6.0
Build log ( ~/.cabal/logs/ghc-parser-0.1.6.0.log ):
[1 of 1] Compiling Main             ( /tmp/ghc-parser-0.1.6.0-6508/ghc-parser-0.1.6.0/dist/setup/setup.hs, /tmp/ghc-parser-0.1.6.0-6508/ghc-parser-0.1.6.0/dist/setup/Main.o )
Linking /tmp/ghc-parser-0.1.6.0-6508/ghc-parser-0.1.6.0/dist/setup/setup ...
unused rules: 20
shift/reduce conflicts:  42
reduce/reduce conflicts: 1
unused rules: 18
shift/reduce conflicts:  46
reduce/reduce conflicts: 1
unused rules: 18
shift/reduce conflicts:  46
reduce/reduce conflicts: 1
Configuring ghc-parser-0.1.6.0...
Warning: 'hs-source-dirs: src-7.10' directory does not exist.
Building ghc-parser-0.1.6.0...
Preprocessing library ghc-parser-0.1.6.0...
setup: can't find source for Language/Haskell/GHC/HappyParser in generic-src,
src-7.10, dist/build/autogen
cabal: Error: some packages failed to install:
ghc-parser-0.1.6.0 failed during the building phase. The exception was:
ExitFailure 1

aueelis:
I noticed this, too.
Although the folders `ghc-7.6`, `ghc-7.8.2`, `ghc-7.8.3` and `ghc-7.10` appear in the repository, only the first one is included in the hackage package.

gibiansky:
Thanks for the note. Looks like `cabal sdist` isn't including `src-7.10` for some reason (see #478). Will fix soon (https://github.com/gibiansky/IHaskell/pull/480)

gibiansky:
This is fixed and will be included in the upcoming Hackage release.

PR #478, opened 2015-04-20
Update for #476, use CPP to manage different versions

sumitsahrawat:
There might be better approaches. Fiddling around with `ghc-parser.cabal` didn't help, so I chose this.

### DIFF
--- a/ghc-parser/ghc-parser.cabal
+++ b/ghc-parser/ghc-parser.cabal
@@ -27,20 +27,14 @@ library
   build-tools:         happy, cpphs
   exposed-modules:     Language.Haskell.GHC.Parser,
                        Language.Haskell.GHC.HappyParser
-  -- other-modules:
+  other-modules:       Language.Haskell.GHC.HappyParser_760
+                     , Language.Haskell.GHC.HappyParser_782
+                     , Language.Haskell.GHC.HappyParser_783
+                     , Language.Haskell.GHC.HappyParser_710
   -- other-extensions:
   build-depends:       base                 >=4.6 && <4.9,
                        ghc                  >=7.6 && <7.11
 
-  if   impl(ghc >= 7.6) && impl(ghc < 7.8)
-    hs-source-dirs:      generic-src src-7.6
-  else
-    if impl(ghc >= 7.8) && impl(ghc < 7.8.3)
-      hs-source-dirs:    generic-src src-7.8.2
-    else
-      if impl(ghc < 7.10)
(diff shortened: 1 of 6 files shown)


gibiansky:
Hmm... This should not be necessary. In https://github.com/gibiansky/IHaskell/issues/476 the poster is trying with GHC 7.10, which *hypothetically* does *not* require a Happy parser build at all, as the GHC 7.10 parser api exposes enough for us to work with. 

The issue is this line:

```
        hs-source-dirs:    generic-src src-7.10
```
There is no auto-generated `src-7.10` directory  (and there shouldn't be; it's not autogenerated). However, it *does* exist in the repo.

I think the required change to #476 is actually adding `src-7.10/Language/Haskell/GHC/HappyParser.hs` to the `extra-source-files` list.

gibiansky:
Implementing the attempt here: https://github.com/gibiansky/IHaskell/pull/480

Closing for now; if #480 doesn't work will re-open and look into CPP. Hopefully can avoid that. Either way thanks for looking into this and kicking this off.

sumitsahrawat:
I experimented with the cabal file, and I noticed something interesting. Reordering the clauses in the if-else chain has visible effects on the contents of the sdist tarball.

For example:

The default order includes `src-7.6` and excludes everything else.

Reordering as follows:

```
  if   impl(ghc >= 7.8) && impl(ghc < 7.8.3)
    hs-source-dirs:      generic-src src-7.8.2
  else
    if impl(ghc >= 7.6) && impl(ghc < 7.8)
      hs-source-dirs:    generic-src src-7.6
    else
      if impl(ghc < 7.10)
        hs-source-dirs:    generic-src src-7.8.3
      else
        hs-source-dirs:    generic-src src-7.10
```

Creates a tarball with 7.8.2 included.

---

Adding `src-7.10/Language/Haskell/GHC/HappyParser.hs` to `extra-source-files` adds that file to the tarball, but there are still files missing from it.

gibiansky:
I went through the same process as well, and was confused as to why it was happening.

Which files are missing from it?

(If you are referring to the `src-7.8.*` directories, then actually that's ok, because the `src-7.6` and `src-7.8` directories shouldn't be in the tarball in the first place. They are generated by `build-parser.sh`, which is called from `Setup.hs` upon installation. So `src-7.6` actually *shouldn't* be in the tarball (and I didn't know until today that it was even being included). *Only* `src-7.10` actually needs to be in the tarball; the other ones are generated from the `.pp` `HappyParser` files.)

PR #488, opened 2015-05-06
7.10.1

FranklinChen:
Got all the displays building on 7.10.1.


gibiansky:
Could you change `.travis.yml` to test the `ihaskell-display` packages on GHC 7.10 as well?

See the line
```
   if [ ${GHCVER%.*} = "7.8" ]; then
```
after `script:`. Change it to
```
   if [ ${GHCVER%.*} = "7.8" ] || [ ${GHCVER%.*} = "7.10" ]; then
```

If that builds it would be awesome to merge this.

Issue #491, opened 2015-05-16
src/IHaskell/Eval/Evaluate.hs:404:16:


marmalodak:
(ihaskell-venv) [root@qa-notebook zeromq-4.0.5]# PKG_CONFIG_PATH=/e2e/lib/zeromq4/lib/pkgconfig:/e2e/lib/zeromq4/lib/pkgconfig:/usr/share/licenses/pkgconfig:/usr/share/pkgconfig:/usr/share/doc/pkgconfig cabal install ihaskell --extra-lib-dirs=/e2e/lib/zeromq4/lib --extra-include-dirs=/e2e/lib/zeromq4/include

...

Loading package MissingH-1.3.0.1 ... linking ... done.
[16 of 18] Compiling IHaskell.Eval.Evaluate ( src/IHaskell/Eval/Evaluate.hs, dist/build/IHaskell/Eval/Evaluate.o )

src/IHaskell/Eval/Evaluate.hs:404:16:
    Couldn't match type `FilePath' with `[Char]'
    Expected type: GHC.IO.FilePath
      Actual type: FilePath
    In the first argument of `writeFile', namely
      `(fpFromString $ directory ++ filename)'
    In a stmt of a 'do' block:
      writeFile (fpFromString $ directory ++ filename) contents
    In the second argument of `($)', namely
      `do { createDirectoryIfMissing True directory;
            writeFile (fpFromString $ directory ++ filename) contents }'

src/IHaskell/Eval/Evaluate.hs:566:40:
    Couldn't match type `FilePath' with `[Char]'
    Expected type: GHC.IO.FilePath
      Actual type: FilePath
    In the return type of a call of `fpFromString'
    In the second argument of `($)', namely `fpFromString filename'
    In a stmt of a 'do' block:
      contents <- readFile $ fpFromString filename
Failed to install ihaskell-0.6.0.0
cabal: Error: some packages failed to install:
ihaskell-0.6.0.0 failed during the building phase. The exception was:
ExitFailure 1

sumitsahrawat:
It seems like this issue arises because of some package upgrade. It installed fine for me, but cabal told me that I should run `cabal update`. Afterwards, reinstalling in a fresh sandbox gives the same error.

gibiansky:
[...]

PR #497, opened 2015-05-22
Temporary fix for classy-prelude

sumitsahrawat:
Add the conversions betwen (String/Text) and FilePath once again.
Will have to be removed again if classy-prelude is removed.

Temporary fix for #491.


gibiansky:
Thanks. This is a serious problem for stability of IHaskell lately, so either a) IHaskell needs to be added to Stackage or b) classy-prelude needs to be removed.

(a) should happen anyways and I will do it soonish (well, next two months) and I think I will also do (b) in the shorter term (next month)

sumitsahrawat:
There's a mistake in here, I'll fix it in a moment.

Issue #495, opened 2015-05-22
Please update ihaskell to recent versions its dependencies


peti:
http://packdeps.haskellers.com/feed?needle=exact%3Aihaskell

We are interested in support for `classy-prelude-0.12.0` in particular, because that dependency is preventing us from building `ihaskell` in NixOS at the moment.

gibiansky:
Does NixOS build from Hackage? Now that IHaskell no longer depends on classy-prelude, do I need to make a new release to make ihaskell nisox-compatible?

peti:
Yes, we build from Hackage. A new release would be very much appreciated.

gibiansky:
Finally done. See [`ihaskell-0.6.4.0`](https://hackage.haskell.org/package/ihaskell).

PR #508, opened 2015-06-02
Backbone widgets

sumitsahrawat:
@gibiansky A lot of changes for you to review.


sumitsahrawat:
There is some code-duplication, which I'll fix with the formatting.

src/IHaskell/Eval/Evaluate.hs
```diff
@@ -321,13 +324,18 @@ evaluate kernelState code output = do
               Just disps -> evalResult evalOut <> disps
           helpStr = evalPager evalOut
 
+      -- Capture all widget messages queued during code execution
+      messagesIO <- extractValue "IHaskell.Eval.Widgets.relayWidgetMessages"
```
sumitsahrawat: Currently, I'm getting an error here, which I've been unable to fix.

```
ihaskell: Not in scope: ‘IHaskell.Eval.Widgets.relayWidgetMessages’
```

sumitsahrawat:
This still requires some work.

sumitsahrawat:
@gibiansky It's an environment problem, just like you said.

Some debug print statements print to the notebook, while other print to the ipython log.

gibiansky:
I've taken a brief look at the code and didn't find anything that jumped out immediately. 

For dealing with environment issues, it's important to remember that the things run by the notebook aren't the things of the main IHaskell codebase, and in fact things like global variables may not be shared. So when you have a NOINLINE'd `TChan`, that `TChan` is not necessarily shared between the two environments.

The way the `displayFromChan` works is that from within the IHaskell environment, you can put stuff into the channel; then, when IHaskell wants to display it, it runs the `unfoldM` in the IHaskell environment to get out all the values, then packages them up, and then transfers them to the compiled environment via `displayFromChan`. It looks like `relayMessages` has been implemented the same way so it seems like the right way to go ahead.

Let me know if this ends up being particularly blocking, and I am happy to take a deeper look and actually play with the code.

sumitsahrawat:
I compared the sequence of messages for the old and new implementation. The sequences differ slightly, as follows

*Both*

* `KernelInfoReply`
* `PublishStatus` (Busy)
* `PublishInput`

*Old implementation only*

* `PublishDisplayData`

```json
{
    "source": "haskell",
    "metadata": null,
    "data": null
}
```

*Both*

* `CommOpen`
* `UpdateState`
* `Display`

* `PublishStatus` (Idle)

*Different in both implementations*

* `ExecuteReply` (Old one shown below)

```javascript
{
  "payload": [ // Empty in the new implementation
    {
      "source": "page",
      "line": 0,
      "data": {
        "text\/plain": ""
      }
    }
  ],
  "execution_count": 1,
  "user_variables": null,
  "user_expressions": null,
  "status": "ok"
}
```

sumitsahrawat:
@gibiansky You wouldn't believe what the issue was. An empty label doesn't display the button, and the label was empty be default.

gibiansky:
@sumitsahrawat Wow, after all that debugging and digging around in internals... Well hopefully it was educational at least. 

Does it work now?

sumitsahrawat:
Yeah, it works great. I'll push an update in a couple of minutes :smile:.

sumitsahrawat:
@gibiansky It works really great now.

Inviting @miguelnegrao to be a beta user

I also created a [minimal working example](https://gist.github.com/sumitsahrawat/d2fdf827df30d7f6111b).

sumitsahrawat:
A remaining issue still is that the click handler is run in the Ghc monad, thus any print output goes into the ipython log. This might require some more nasty hacks.

gibiansky:
@sumitsahrawat Great, looks awesome. The remaining issue definitely needs to be dealt with, though. Have you tried this?

```haskell
ref <- newIORef 0
onClicked b $ const $ do
  modifyIORef' ref (+1)
  readIORef ref >>= (setButtonLabel b . show)
```

Does that work?

Also, can you change the order of parameters, so that `b` goes first, e.g.:

```haskell
setButtonLabel :: Button -> String -> IO ()
```
This mirrors a lot of common interfaces (see Data.IORef, for example) and allows currying as used in the small snippet above.

Anyway those are minor things. Great work! Once the click handler runs properly in the interpreted environment, I can do a full code review, nitpick a bunch of things, then merge it!

If your work continues at this pace you're going to get a ton more done this summer than I expected :)

gibiansky:
And yes, dealing with environments in IHaskell has pretty much always needed incredibly nasty hacks, so if that's what it ends up needing, go ahead. Just be very sure to document the nastiness and hackiness of the code in comments – multiparagraph comments explaining why hacks are necessary are perfectly ok if the hacks need that sort of explanation.

sumitsahrawat:
> Also, can you change the order of parameters, so that b goes first

I agree completely. I found myself confused sometimes, and this seems better.

> Does that work?

The example works, even though it requires explicit packing from `String` to `Text`. Shall I switch to using `String` by default?

> If your work continues at this pace you're going to get a ton more done this summer than I expected :)

I hope so :smile:

gibiansky:
I'm not sure about `String` vs `Text`, actually. I was debating that as well when writing it.

The main issue with `Text` is that it requires users to turn on `OverloadedStrings` to use easily, but that's not really a big deal. 

For now leave it as `Text` if that's what it is; and let's default to `Text` for other similar things in the future in the widgets API. After a few months of usage we can see whether it gets in the way, and if necessary, change it to `String`.

sumitsahrawat:
Still some minor issues.

- [ ] Prevent kernel crash on `on_click` causing exceptions/`error`.
- [ ] Add functions to close buttons.
- [ ] Change `getButtonText` to `getButtonLabel`

Will get to these asap.

sumitsahrawat:
@gibiansky It still freezes if the widget demands console input. IPython crashes if a button demands input (in python).

We might be able to do better, but I'll leave it for later.

gibiansky:
So the previous issues with `putStrLn` and so on were dealt with? Awesome! Can you describe what you did and why it worked? (I haven't read the code but if there are confusing hacky sections make sure to document them. But this will come up during code review anyways.)

The todos are good; make sure also that Travis CI is building properly. If possible I would like to do the code review and merge once the button widget is complete (rather than wait for other widgets). After that it will be easier to merge new widgets as they involve less changes to core IHaskell.

Also it'd be great if you could file a bug in IPython about the failing on stdin (but make sure that one hasn't been filed already, of course).

sumitsahrawat:
I did file an issue ipython/ipywidgets#44.

I'll have to look more closely to determine why this worked. I tried it in many different ways (this was the third branch), and this one worked.

Should I catch more exceptions in `IHaskell.Eval.Evaluate.capturedIO`? I have currently set up catch for `ErrorCall` which handles both `error` and `undefined` within click handlers.

gibiansky:
Ah, alright. 

The fundamental rule is that no Haskell code you write in IHaskell should be able to *crash* IHaskell. So *all* exceptions thrown by widget event handlers should be caught *somehow* – they should never crash the kernel. Use `SomeException` would be my suggestion, I think. In what cases would you not want them to be handled as such? 

What happens right now if you enter `error "XYZ"` as the button handler? Does the error message appear anywhere, and, if so, where?

sumitsahrawat:
It goes to the notebook using `displayError`.

I didn't want to use `SomeException` because its use is discouraged, even more when the user can hit Ctrl-C to cause an exception anytime.

sumitsahrawat:
Switched to using `SomeException`

sumitsahrawat:
@gibiansky I'd recommend making a new release for #495 before merging this.

gibiansky:
Agreed. Is this ready to code review / merge?

sumitsahrawat:
Yes. :+1:

ihaskell-display/ihaskell-aeson/IHaskell/Display/Aeson.hs
```diff
@@ -16,4 +16,4 @@ instance IHaskellDisplay Value where
   display renderable = return $ Display [plain json, html dom]
     where
       json = T.unpack $ E.decodeUtf8 $ LBS.toStrict $ encodePretty renderable
-      dom = [i|<div class="highlight-code" id="javascript">${json}</div>|]
+      dom = [i|<div class="highlight-code" id="javascript">${json}</div>|]
```
gibiansky: Can you add a newline to the end of these files? It creates unnecessary diff changes. (`hindent` skips that newline, but generally I prefer to have that newline. I use `hindent` from inside an editor, which avoids that, but you could also just re-add that newline after piping stuff through hindent.)

sumitsahrawat: I'll go ahead and checkout the unnecessary changes.

ihaskell-display/ihaskell-widgets/src/IHaskell/Display/Widgets/Button.hs
```diff
@@ -0,0 +1,210 @@
...
+    Button,
+    -- * Predefined button styles
+    ButtonStyle(..),
+    -- * Create a new button
+    mkButton,
+    -- * Set button properties
+    setButtonStyle,
+    setButtonLabel,
+    setButtonTooltip,
+    disableButton,
```
gibiansky: It seems more consistent to use `set`, as in `setButtonEnabled`. So instead of `enableButton`, `disableButton`, and `isDisabled`, switch to `setButtonEnabled` and `getButtonEnabled`.

Not for now, but when we build a bigger API on top of this, maybe considering creating lenses for these and exporting them would work somehow.

ihaskell-display/ihaskell-widgets/src/IHaskell/Display/Widgets/Button.hs
```diff
@@ -0,0 +1,210 @@
...
+import           Data.Text (Text)
+import qualified Data.Text as T
+import           System.IO.Unsafe (unsafePerformIO)
+
+import           IHaskell.Display
+import           IHaskell.Eval.Widgets
+import qualified IHaskell.IPython.Message.UUID as U
+import           IHaskell.Types (WidgetMethod(..))
+
+-- | ADT for a button
```
gibiansky: Somewhat unnecessary comment :)

This would be a good place to explain how buttons are used (what their capabilities are and what functions access those capabilities).

ihaskell-display/ihaskell-widgets/src/IHaskell/Display/Widgets/Button.hs
```diff
@@ -0,0 +1,210 @@
...
+data Button =
+       Button
+         { uuid :: U.UUID
```
gibiansky: Why is this UUID here? Explain briefly.

If this is a pattern that will occur across all the widgets, then it may make sense to wrap this somehow so that all widgets have the same field and use the same way to access that field. That may be refactored later, though.

ihaskell-display/ihaskell-widgets/src/IHaskell/Display/Widgets/Button.hs
```diff
@@ -0,0 +1,210 @@
...
+         , description :: IORef Text
```
gibiansky: An explanation of where each of these texts and bools and so on are used would be nice. Very brief, but just like "text inside the button" or "shown on hover", etc.

ihaskell-display/ihaskell-widgets/src/IHaskell/Display/Widgets/Button.hs
```diff
@@ -0,0 +1,210 @@
...
+  -- Default properties, with a random uuid
+  uuid <- U.random
+  sender <- newIORef Nothing
+  desc <- newIORef "label"        -- Non-empty to get a display
+  ttip <- newIORef ""
+  dis <- newIORef False
+  sty <- newIORef None
+  fun <- newIORef (\_ -> return ())
+
+  let b = Button uuid desc ttip dis sty fun
```
gibiansky: For a complex thing such as a button, position independent record syntax may be preferable.

ihaskell-display/ihaskell-widgets/src/IHaskell/Display/Widgets/Button.hs
```diff
@@ -0,0 +1,210 @@
...
+                 | Danger
+                 | None
+  deriving (Eq, Show)
+
+-- | Create a new button
+mkButton :: IO Button
+mkButton = do
+  -- Default properties, with a random uuid
+  uuid <- U.random
+  sender <- newIORef Nothing
```
gibiansky: What is this `sender`?

ihaskell-display/ihaskell-widgets/src/IHaskell/Display/Widgets/Button.hs
```diff
@@ -0,0 +1,210 @@
...
+
+  let b = Button uuid desc ttip dis sty fun
+
+  -- Open a comm for this widget, and store it in the kernel state
+  widgetSendOpen b (toJSON ButtonInitData) (toJSON b)
+
+  -- Return the button widget
+  return b
+
+update :: Button -> [Pair] -> IO ()
```
gibiansky: What do `update` and `modify` do? What is the difference between them?

What is a `Pair` and why do we take a list of them?

gibiansky: If `Pair` is from `aeson`, perhaps you could use `FromJSON` directly in these rather than forcing all their users to call `toJSON`?

gibiansky:
Not quite done, but have to go for now. Will continue code review later.

sumitsahrawat:
I'll fix all the issues you point out tomorrow. All in one go.

All the widgets have elements from the same set of properties, most of which are common for all widgets. I'll refactor the button implementation so that the widget agnostic parts live in one place.

ihaskell-display/ihaskell-widgets/src/IHaskell/Display/Widgets/Button.hs
sumitsahrawat: All widgets need to store the UUID of their comm. I enforced this by adding the `getCommUUID` function to `IHaskellWidget`.

ihaskell-display/ihaskell-widgets/src/IHaskell/Display/Widgets/Button.hs
sumitsahrawat: Artifact from an old design, sorry :sweat_smile:

ihaskell-display/ihaskell-widgets/src/IHaskell/Display/Widgets/Button.hs
sumitsahrawat: > What do update and modify do? What is the difference between them?
> 
> What is a Pair and why do we take a list of them?

For example,

```haskell
-- | Set the button tooltip
setButtonTooltip :: Button -> Text -> IO ()
setButtonTooltip b txt = do
  modify b tooltip txt  -- Put 'txt' in the 'tooltip' IORef from 'b'
  update b ["tooltip" .= txt]
  -- Update the button stored in the kernelState, and only send
  -- the modification ({ tooltip: txt }) to the notebook
```

sumitsahrawat:
@gibiansky Take a look at [this notebook](http://nbviewer.ipython.org/gist/sumitsahrawat/e3bd5c2c1f492f9e41c2), where I experimented with all the properties the button widget has.

It would be better to provide all the possible properties separately, and define a widget as a collection of those properties. This way, `ToJSON` coulld be defined separately for all properties, which would make it easier to combine `update` and `modify` as seen above.

sumitsahrawat:
@gibiansky I'm trying to implement a `WidgetModel`, which will correspond with the generic model from IPython frontend. There is a lot of repitition, which I'm trying to solve using template haskell.

When it's done, it'll make it very easy to implement other widgets. Once I have a scheme which I'm sure won't see many changes, I'll make this ready to merge.

gibiansky:
I've generally tried to avoid template haskell when possible. Why do you need it here?

Can you instead just have two data types, composed into one, via something like this?
```haskell
data Widget a = Widget { ... , getWidget :: a }
data ButtonData = ButtonData { description :: String, ... }
type Button = Widget ButtonData
```
Or something along this path, rather than generating code via TH?

sumitsahrawat:
I'm trying to generalize the widget representation, so that all widgets use the same data structure. The following is a rough sketch:

```haskell
module Attributes where

class Attribute a where
  pairs :: a -> [Pair]

-- Requires UndecidableInstances, but can be fixed using a newtype
instance Attribute a => ToJSON a where
  toJSON = object . pairs

-- Some attributes. All attributes can be converted to Pairs,
-- which can then be made into a big object using `object . concat`,
-- while also providing toJSON functionality individually.

-- For example:

data ViewName = ViewName String

instance Attribute ViewName where
  pairs (ViewName name) = ["_view_name" .= str name]

-------------------------

module WidgetModel where

data WidgetModel = WidgetModel UUID MutableAttributes

-- All possible attributes.
data MutableAttributes = MutableAttributes { _attr :: Maybe (<some attribute>) }

-- Skip widget attributes with `Nothing` values
widgetAttributeToPairs :: WidgetAttribute -> [Pair]
widgetAttributeToPairs (WidgetAttribute m) = maybe [] pairs m
```

I needed template haskell to generate small functions like these:

```haskell
viewName = WidgetAttribute . _viewName
```

gibiansky:
Perhaps the widget can instead be represented by some sort of `Map`? Instead of as a record.

In particular I am somewhat hesitant to do template haskell immediately. It seems like overarchitecting. Let's make 2-3 widgets first, and _then_ try to generalize everything.

I'll continue to code review this so we can merge it soon. Will also make a release soon.

sumitsahrawat:
Yeah. I'll keep experimenting in the meantime.

Also, I'm thinking of working on a selection widget next, maybe dropdown menus.

gibiansky:
Good plan.

ihaskell-display/ihaskell-widgets/src/IHaskell/Display/Widgets/Button.hs
```diff
@@ -0,0 +1,213 @@
...
+-- | Create a new button
+mkButton :: IO Button
+mkButton = do
+  -- Default properties, with a random uuid
+  commUUID <- U.random
+  desc <- newIORef "label"      -- Non-empty to get a display
+  ttip <- newIORef ""
+  dis <- newIORef False
+  sty <- newIORef None
+  fun <- newIORef (\_ -> return ())
```
gibiansky: `newIORef $ const $ return ()`

main/Main.hs
```diff
@@ -362,21 +293,40 @@ replyTo _ HistoryRequest{} replyHeader state = do
         }
   return (state, reply)
 
-handleComm :: (Message -> IO ()) -> KernelState -> Message -> MessageHeader -> IO KernelState
-handleComm replier kernelState req replyHeader = do
+handleComm :: (Message -> IO ()) -> KernelState -> Message -> MessageHeader -> Interpreter KernelState
```
gibiansky: This is pretty dense code, even if it is pretty clear. Could you describe what `handleComm` does (in a comment) and maybe add comments in the function itself if there is anything you think is tricky and needs explaining?

src/IHaskell/Eval/Evaluate.hs
```diff
@@ -335,12 +344,25 @@ evaluate kernelState code output = do
 
     storeItCommand execCount = Statement $ printf "let it%d = it" execCount
 
-    extractValue :: Typeable a => String -> Interpreter a
-    extractValue expr = do
-      compiled <- dynCompileExpr expr
-      case fromDynamic compiled of
-        Nothing     -> error "Error casting types in Evaluate.hs"
-        Just result -> return result
+extractValue :: Typeable a => String -> Interpreter a
```
gibiansky: Add a haddock comment explaining this

src/IHaskell/Eval/Evaluate.hs
```diff
@@ -1093,10 +1052,10 @@ keepingItVariable act = do
   goStmt $ printf "let it = %s" itVariable
   act
 
-capturedStatement :: (String -> IO ())         -- ^ Function used to publish intermediate output.
-                  -> String                            -- ^ Statement to evaluate.
-                  -> Interpreter (String, RunResult)   -- ^ Return the output and result.
-capturedStatement output stmt = do
+capturedEval :: (String -> IO ()) -- ^ Function used to publish intermediate output.
+             -> Either String (IO a) -- ^ Statement to evaluate.
```
gibiansky: Replace `Either` with a custom data type:
```haskell
data CapturedEval a = CapturedStmt String | CapturedIO (IO a)
```
Otherwise this looks like the classic `Either` with an error message `String` attached to it.

src/IHaskell/Eval/Widgets.hs
```diff
@@ -0,0 +1,145 @@
...
+          -- Initial state update
+          communicate . toJSON $ UpdateState stateVal
+
+          -- Send anything else the widget requires.
+          open widget communicate
+
+          -- Store the widget in the kernelState
+          return newState
+
+    (Close widget value) -> do
```
gibiansky: Parens not necessary; here and in a few other places. Make sure you run `hlint` on all of the source files, it points out silly things like that, and I like to keep `hlint` happy (some of its suggestions are a bit pointless but it keeps coding style consistent).

src/IHaskell/Eval/Widgets.hs
```diff
@@ -0,0 +1,145 @@
...
+widgetSendUpdate :: IHaskellWidget a => a -> Value -> IO ()
+widgetSendUpdate = widgetSend Update
+
+widgetSendView :: IHaskellWidget a => a -> IO ()
+widgetSendView = queue . View . Widget
+
+widgetSendClose :: IHaskellWidget a => a -> Value -> IO ()
+widgetSendClose = widgetSend Close
+
+handleMessage :: (Message -> IO ())
```
gibiansky: This functions is quite long and complex. Could you add a top-level comment explaining what it is? Also, perhaps it would be worth breaking into some smaller functions to make individual bits more understandable, not sure.

src/IHaskell/Eval/Widgets.hs
```diff
@@ -0,0 +1,145 @@
...
+          newState = state { openComms = newComms }
+
+      if present
+        then do
+          header <- dupHeader replyHeader CommDataMessage
+          send . CommData header uuid . toJSON $ UpdateState value
+          return newState
+        else return state
+
+widgetHandler :: (Message -> IO ())
```
gibiansky: Is there perhaps a cleaner way to write this? It looks like a `fold` but monadic (`foldM` from `Control.Monad`, or something along those lines?)

src/IHaskell/Publish.hs
```diff
@@ -0,0 +1,76 @@
...
+import           IHaskell.Display
+import           IHaskell.Types
+
+ihaskellCSS :: String
+ihaskellCSS = [hereFile|html/custom.css|]
+
+-- Publish outputs, ignore any CommMsgs
+publishResult :: (Message -> IO ())
+              -> MessageHeader
+              -> MVar [Display]
```
gibiansky: What do each of these arguments mean?

src/IHaskell/Types.hs
```diff
@@ -56,25 +65,32 @@ class IHaskellDisplay a where
 
 -- | Display as an interactive widget.
 class IHaskellDisplay a => IHaskellWidget a where
-  -- | Output target name for this widget. The actual input parameter should be ignored.
+  -- | Output target name for this widget. The actual input parameter should be ignored. By default
+  -- evaluate to "ipython.widget", which is used by IPython for its backbone widgets.
   targetName :: a -> String
+  targetName _ = "ipython.widget"
```
gibiansky: Can we change this to use `Proxy` from `Data.Proxy` instead of ignoring the function parameter? (Maybe not, if it poses a 7.6.3 problem, but we should try.)

gibiansky:
Alright, reviewed the rest of the code. Looks pretty clean and solid; only minor comments all-around. Once those are addressed and the Travis CI build passes I think it'll be ready for merge. Woohoo!

sumitsahrawat:
Thanks, I'll get to it now. Expect it to be done in a couple of hours max.

src/IHaskell/Types.hs
sumitsahrawat: I read up on `Data.Proxy` and finally understood what it was. Then I tried to use it, and it seems that it won't work here. It causes problems with defining an `IHaskellWidget` instance for `Widget`.

```haskell
targetName (Widget w) = targetName w -- Current definition
targetName (Proxy :: Widget) = targetName (Proxy :: ???)
```

sumitsahrawat:
@gibiansky Does it make sense to call it `ButtonWidget` instead of `Button`. IPython has dropped the `Widget` suffix, whereas we will have to use it *atleast* for `TextWidget`.

gibiansky:
@sumitsahrawat Is this ready to merge (other than the refactoring we'll discuss after it's merged)?

sumitsahrawat:
Yes, it looks ready to merge. :+1:

sumitsahrawat:
@gibiansky Adding `ImageWidget`, which is also ready.

sumitsahrawat:
@gibiansky The `OutputWidget` is coming :smile:

sumitsahrawat:
Ran into some issues with `OutputWidget`. It allows appending of outputs, but sadly, clearing output doesn't work as expected. Will have to investigate further.

Feel free to merge this in the meantime.

gibiansky:
@sumitsahrawat Great, will merge as soon as last commit is verified on Travis!

sumitsahrawat:
Made a mistake while separating `output-widget` and `widget` branches. Fixed.

gibiansky:
It's late here right now so I'm off to sleep. I'll merge this tomorrow morning and will give some ideas on how to refactor it as well.

The main idea I have in mind is to change from this
```haskell
data Button =
       Button
         { uuid :: U.UUID       -- ^ The UUID for the comm
         , description :: IORef Text -- ^ The label displayed on the button
         , tooltip :: IORef Text     -- ^ The tooltip shown on mouseover
         , disabled :: IORef Bool    -- ^ Whether the button is disabled
         , buttonStyle :: IORef ButtonStyle -- ^ The button_style
         , clickHandler :: IORef (Button -> IO ()) -- ^ Function executed when button is clicked
         }

-- | Mechanism for updating the actual in-memory widget.
modify :: IHaskellWidget a => a -> (a -> IORef b) -> b -> IO ()

-- | Mechanism for updating the frontend by sending a JSON blob.
update :: IHaskellWidget a => a -> [Pair] -> IO ()
```
to perhaps using a heterogenous typed list slash record from Vinyl, using their rough idea of a lens for accessing/setting:
```haskell
import Data.Vinyl

-- Possible widget fields declared as Vinyl record fields...
description = Field ::: "description" ::: IORef Text
clickHandler = Field ::: "click_handler" ::: IORef (Button -> IO ())

-- Fields for a button
type ButtonFields = [description, clickHandler]

-- | ADT for *any* widget
data Widget f = Widget { uuid :: UUID, fields :: f }

-- | Type for a button
type Button = Widget ButtonFields

-- | modify and update combined
change :: Widget f -> sing r -> r -> IO ()
change = use rput?
```

(Will investigate more later.) If this proves to be too complex, then we can just deal with code duplication.

miguel-negrao:
> @gibiansky The OutputWidget is coming :smile:

Unfortunately I don't have time at the moment to test any of this, but I'm really looking forward for the final result ! :-) If you take requests, a slider widget would be very useful for all sorts of things, such as math and graph visualizations, since it basically represents a range of real numbers.

sumitsahrawat:
@miguel-negrao Sliders have a big hierarchy in IPython. I actually started the implementation with sliders, but switched to buttons due to their simplicity.

I'll implement them next, but after some refactoring of the current widgets. Will be sure to ping you then. Thanks for chiming in.

PR #514, opened 2015-06-15
Fix reading png-files with bytestring

laevar:
When testing the ihaskell-diagrams backend with the sierpinski-triangle i got

.ihaskell-diagram.png: hGetContents: invalid argument (invalid byte sequence)

in Diagrams.hs the readFile is 

       filePath -> IO String 

and thus couldn't read the Bytestring png. I used the Char.readFile which is 

       filePath -> IO Bytestring

This would break svg, as it needs String, thus the Byestring will be unpacked before using svg.


gibiansky:
Caught in #517, I believe. Thanks! (If these two PRs are actually different and I'm misunderstanding, let me know.)

PR #519, opened 2015-06-28
Refactoring Widgets + Boolean Widgets

sumitsahrawat:
Still have a long way to go.


ihaskell-display/ihaskell-widgets/src/IHaskell/Display/Widgets/Types.hs
```diff
@@ -0,0 +1,204 @@
...
+setField widget proxy val = do
+  let field = reflect proxy
+  modifyIORef (attrs widget) (M.insert field $ toDyn val)
+
+getField :: (Typeable (FieldType f), SingI f, IsElem (WidgetFields w) f ~ True) => Widget w -> Proxy f -> IO (FieldType f)
+getField widget proxy = do
+  let field = reflect proxy
+  valMap <- readIORef $ attrs widget
+  case M.lookup field valMap of
+    Nothing -> return undefined
```
gibiansky: Use `error "something"` instead. Is it supposedly impossible for this to happen?

sumitsahrawat: It *should* not happen, if the widgets were constructed properly and the properties were not modified. It will hopefully be well hidden to prevent such editing.

gibiansky:
@sumitsahrawat How is the port to `vinyl` going?

sumitsahrawat:
@gibiansky The last commit results in somewhat usable widgets. The issue was with default properties.

The trouble with numerical values is that IPython represents them as strings, and the values are kept within acceptable ranges using the traitlets machinery. Any property which is sent as an empty string during initial state update automatically gets the default value on the frontend side. To do this, I was using a newtype (`PosInt`), and it worked transparently.

Using the same newtype with vinyl doesn't work transparently. For example, previously:

```haskell
setButtonWidth b 400 -- Passing an Int
```

And, now with vinyl:

```haskell
setField b SWidth (PosInt 400) -- Explicitly pass a PosInt
```

I don't think there's an easy way to get around this. Otherwise, the port is complete :tada:

sumitsahrawat:
@gibiansky Added boolean widgets, i.e. `CheckBox` and `ToggleButton`.

sumitsahrawat:
@gibiansky The latest commit should pass the build, except for hindent which errors out.

```
hindent: FIXME: No implementation for TyPromoted.
```

sumitsahrawat:
This is strange. The build passed for me in a fresh sandbox using ghc 7.8.4. I'll look into it later.

gibiansky:
I'm impressed by how little code it takes to add a widget. Awesome!

Ignore the `hindent` issues. Other than that this is ready to merge?

sumitsahrawat:
@gibiansky There seems to be some issue with ghc-7.8.4. I tried to recreate it by downgrading my ghc but couldn't determine what it was. I'll look into it tomorrow.

sumitsahrawat:
@gibiansky This is ready to merge, ignoring `hindent`.

ihaskell-display/ihaskell-widgets/ihaskell-widgets.cabal
```diff
@@ -56,24 +56,31 @@ library
   -- Modules included in this library but not exported.
   other-modules:       IHaskell.Display.Widgets.Button
                        IHaskell.Display.Widgets.Image
-                       IHaskell.Display.Widgets.Dropdown
+                       IHaskell.Display.Widgets.Bool.CheckBox
+                       IHaskell.Display.Widgets.Bool.ToggleButton
+                       -- IHaskell.Display.Widgets.Dropdown
```
gibiansky: Is this unimplemented? If so, it should be removed prior to merge from module list, otherwise, uncommented.

ihaskell-display/ihaskell-widgets/ihaskell-widgets.cabal
```diff
@@ -84,3 +91,6 @@ library
   -- Base language which the package is written in.
   default-language:    Haskell2010
   
+  -- Deal with small -fcontext-stack on ghc-7.8
```
gibiansky: Does this work on GHC 7.6, then?

ihaskell-display/ihaskell-widgets/src/IHaskell/Display/Widgets/Types.hs
```diff
@@ -0,0 +1,249 @@
...
+-- All records with ToPair instances for their Attrs will automatically have a toJSON instance now.
+instance RecAll Attr (WidgetFields w) ToPairs => ToJSON (WidgetState w) where
+  toJSON record =
+    object
+    . concat
+    . recordToList
+    . rmap (\(Compose (Dict x)) -> Const $ toPairs x)
+    $ reifyConstraint (Proxy :: Proxy ToPairs) $ _getState record
+
+data Widget (w :: WidgetType) = Widget { uuid :: UUID, state :: IORef (WidgetState w) }
```
gibiansky: We already have something called `Widget`; we should avoid using the same name in different places. This creates confusion and requires import with `hiding`. So I would probably vote to rename this one rather than the old one; perhaps change this to `BuiltinWidget` or `UIWidget` really anything other than just `Widget`.

ihaskell-display/ihaskell-widgets/src/IHaskell/Display/Widgets/Types.hs
```diff
@@ -0,0 +1,249 @@
...
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE QuasiQuotes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE NoMonomorphismRestriction #-}
+{-# LANGUAGE TypeSynonymInstances #-}
+{-# LANGUAGE BangPatterns #-}
+{-# LANGUAGE ConstraintKinds #-}
+module IHaskell.Display.Widgets.Types where
```
gibiansky: This module seems like the real meat of the implementation.

Can you add some documentation for how this works? Specifically, what are the relevant `data` clauses, type families, and so on? How are singletons used? Basically, if I am someone who has never read this codebase before, understands all the extensions used, and is only marginally familiar with singletons, what do I need to know in order to understand the overall design and the specifics? (Basically a long comment at the top with architecture details and a comment on top of each `data` and `type family` describing their role in the architecture)

ihaskell-display/ihaskell-widgets/ihaskell-widgets.cabal
sumitsahrawat: Yup, ghc-7.6 had a default value of 200, it seems.

ihaskell-display/ihaskell-widgets/ihaskell-widgets.cabal
sumitsahrawat: I'll try and get the selection widget working in this PR then. I ran into some issues when trying to implement them in full generality.

gibiansky:
Besides the small numbers of comments I left already, the code looks good. I'd like to see a little bit more effort into documenting how it works (see my comment about that earlier). 

In addition to documenting the type-level architecture used, I'd really like a long section of documentation about *how* the protocol works. Specifically, what messages do the widgets send, in what order, which are needed for initializing widgets, what is the role of default values (why do we use a lot of ""), do widgets ever get destroyed, what fields do those messages have, etc etc? You can reverse engineer the super-precise details (field *names* and values) from the code, but you shouldn't need to reverse engineer the architecture from that. I'm not sure which module this documentation should best go in but it should be somewhere in one of the main widget modules.

So, pending the selection widget, a few top-level comments on each of the data declarations / type synonyms / type families / type classes, and those two large chunks of top-level documentation (type-level architecture and IPython widget protocol), looks good to go.

ihaskell-display/ihaskell-widgets/src/IHaskell/Display/Widgets/Types.hs
sumitsahrawat: I'lll rename it to `BackboneWidget` then. It'll be more convenient that way.

sumitsahrawat:
Perfect :+1:. I'll resolve the shortcomings in a couple of days max.

gibiansky:
Great. I'd love to get this merged by Tuesday or Wednesday.

What widgets remain on the horizon after this change goes through, and how much effort do you anticipate their implementation being?

sumitsahrawat:
Part of the widget heirarchy to be added in this PR:

- [ ] _Selection
  - [ ] Dropdown
  - [ ] RadioButtons
  - [ ] Select
  - [ ] ToggleButtons
  - [ ] _MultipleSelection
    - [ ] SelectMultiple

What remains to be added afterwards:

- Box
  - FlexBox
  - _SelectionContainer
    - Accordion
    - Tab

And the `Num` widgets, where `Num` is either `Int` or `Float`.

- _Num
  - NumText
  - _BoundedNum
    - BoundedNumText
    - NumProgress
    - NumSlider
  - _BoundedNumRange
    - NumRangeSlider

sumitsahrawat:
The `Num` widgets should be easy to implement. Just like the `String` widgets.

When implementing the Box widgets, we'll have the chance to experiment with parent/children widgets, i.e a widget inside another widget. That opens up the possibility of `interact`, while also adding a lot of  complexity. Thus, I can't say how long it'll take to implement them in full generality.

gibiansky:
Great, sounds reasonable. Those can be separate PR s after this merge.

ihaskell-display/ihaskell-widgets/src/IHaskell/Display/Widgets/Types.hs
sumitsahrawat: @gibiansky `IPythonWidget` look good?

gibiansky: Sure.

sumitsahrawat:
@gibiansky Added rest of the widgets. Also made hindent ignore the complicated source files.

sumitsahrawat:
The latest commit makes some changes to the `build.sh` file. The `ihaskell-widgets` package seems to take some time during the build, so I've made it install separately.

gibiansky:
A monumental and awesome PR. Way to go!

sumitsahrawat:
Thanks :+1:

PR #529, opened 2015-07-17
Add support for stack

pavelkogan:


gibiansky:
This is great! Could you add a blurb to the README describing how to install IHaskell with `stack` (and perhaps a link to instructions on how to install stack)? I've been trying to keep the readme complete, since IHaskell may be an entry point for new Haskell users.

pavelkogan:
OK, I've added instructions for installing with `stack`, with a link to its readme.

gibiansky:
Awesome! Guess it's time to finally try `stack` then, been meaning to for a while now.

PR #534, opened 2015-07-21
Add withDynamic to IPython.Kernel.ZeroMQ

joehendrix:
Hi,

I have been using the IPython Kernel bindings in an internal Galois symbolic execution.  For historical reasons, our REPL uses these bindings even when running in a standard command line mode.

This pull request introduces a new primitive that allows the IPython bindings to be setup with arbitrary localhost ports, thus allowing us to run multiple instances at once.  It does this by introducing a new function `withDynamic` that creates a new ZeroMQ interface to a callback along with a `ZeroMQDynamicPorts` that describes the ports chosen.

Let me know if you have any questions, I'd be happy to fix design issues, but it seems like this would be useful to add to the current IPython kernel bindings.

### DIFF
--- a/ipython-kernel/src/IHaskell/IPython/ZeroMQ.hs
+++ b/ipython-kernel/src/IHaskell/IPython/ZeroMQ.hs
@@ -10,8 +10,8 @@ module IHaskell.IPython.ZeroMQ
        , ZeroMQStdin(..)
        , serveProfile
        , serveStdin
-       , ZeroMQDynamicPorts(..)
-       , withDynamic
+       , ZeroMQEphemeralPorts(..)
+       , withEphemeralPorts
        ) where
 
 import           Control.Concurrent
@@ -26,7 +26,8 @@ import           Data.Digest.Pure.SHA as SHA
 import           Data.Monoid ((<>))
 import qualified Data.Text.Encoding as Text
 import           System.IO.Unsafe
-import           System.ZMQ4 hiding (stdin)
+import           System.ZMQ4 as ZMQ4 hiding (stdin)
+import           Text.Read (readMaybe)
 
 import           IHaskell.IPython.Types
 import           IHaskell.IPython.Message.Parser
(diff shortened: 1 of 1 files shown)


gibiansky:
I'm fairly open to this sort of contribution and am happy to work with you to get it merged. There's a few very minor style issues, but I also don't actually understand what this is trying to do. Can't you use the `serveProfile` function with a custom `Profile` to do this already?

joehendrix:
The serveProfile function expects a Profile object, which defines the ports to use.  If those ports are already in use, then you the the calls to bind in serveSocket fail in a separate thread.

The "withDynamic" function in contrast only requires the HMAC key and debug Boolean fields from Profile.  The calls to ZeroMQ bind allow arbitrary ports to be chosen.

Our applications startup process is different from how I think typical uses go, so perhaps we should revisit it.  In any case, it is:

 * Start Haskell program.
 * Create zeroMQ ports to talk to iPython (using withDynamic).
 * Launch iPython with "--existing" flag to telling ipython which zeroMQ ports to talk to.
 * Communicate with ipython via the ZeroMQInterface returned by withDynamic.

With this approach, we don't have an existing profile when creating the ZeroMQ ports; an early version was hardcoding them, but this resulted in error when multiple instances of our application were running concurrently.

This is backwards from how ipython typically is launched since in this case the kernel is starting ipython rather than the other way around; it provides users with a consistent interface.  We considered having our process relaunch itself as a kernel, but this seemed a bit cleaner.

gibiansky:
Could we merge `withDynamic` and `serveProfile` by wrapping each port in a `Maybe`? Or perhaps the entire profile?

Then `serveProfile` gets the type
```haskell
serveProfile :: Profile            -- ^ The profile specifying which ports and transport mechanisms to use.
             -> Bool               -- ^ Print debug output
             -> IO (Profile, ZeroMQInterface)
```
where for any port that is `Nothing` instead of `Just int` it binds a dynamic port.

Is that a possible refactor, to avoid the duplication we currently see behind `withDynamic` and `serveProfile`?

joehendrix:
I just changed withDynamic to withEphemeralPorts to more accurately reflect what it is doing.  I agree dynamic is heavily overloaded.

joehendrix:
The read was changed to readMaybe, and the IO Monad's instance of fail is called with a detailed error message if the pattern match fails.  This should never actually occur in a real execution (a TCP port should be parsable as an integer), but if it does at least the error message will make tracking the bug down easier.

joehendrix:
I find the existing serveProfile fairly dangerous.  In particular, it runs the ZeroMQ bind commands in separate threads.  If those bind commands raise exceptions, they are not passed back to the original calling thread, and so it is difficult to detect.

I tend to think that serveProfile should be dropped and a "withProfile" type function added, but I don't know if that would address all the use cases for serveProfile.

I also think it would be best to find a way to avoid spawning new threads in withEphemeralPorts, but that's a bigger refactoring as they are essential to carrying data between the Channels in ZeroMQInterface and the ZeroMQ sockets.  One approach may be to rewrite the ZeroMQInterface to export callback functions rather than the Control.Concurrent.Chan channels.

gibiansky:
Good points. As far as I know, `serveProfile` is only used in `main` and similar use cases. There should be no negative repercussions to moving all of the `bind` into a the main thread and only doing the actions in the separate threads.

What would `withProfile` do differently from a fixed version of `serveProfile`?

I think I agree with you on the `Chan` vs callback issue but it's not a pressing refactoring – we haven't had any problems with that design decision so far and so while I agree with the idea it's not very pressing to refactor that.

joehendrix:
The interface for `serveProfile` serve profile definitely requires a separate thread is launched for the call to `System.ZMQ4.withContext`.

To get the bind exceptions back to the original thread calling `serveProfile`, it seems like one would then need to either: (1) run the bind calls in that thread and marshall any exceptions back to the thread calling `serveProfile`, or (2) marshall the context back to the main thread and do the bind calls.  I think (1) would be simpler, but it is still a fair bit of work.

I personally don't think think there is too much redundancy between these two interfaces, but I'm also not the maintaining it.  :)  I think most users are going to want to specify either all the ports or none of them, so the extra flexibility of having a `Maybe` type around each port in `Profile` may not be needed.  Also I'd prefer not to break any existing code calling `serveProfile`.

joehendrix:
I tried to modify the function to end the threads spawned by `withEphemeralPorts` but ran into some asynchronous exceptions that I couldn't find an easy workaround for.  I think the threads will need to stay.

I'll fix minor things, but I'm not sure how much time I have for developing a unified interface with `serveProfile`.  If there's a concern this is duplicative of `serveProfile`, I think it's just easiest for us to maintain a public fork for now.

gibiansky:
Nah, I think I am fairly convinced that unifying these two interfaces is tricky. The threading seems difficult and I may unify them later if I end up refactoring this part of the codebase to use callbacks.

Once Travis CI is done I'll check whether the failure is due to a bad setup (right now everything fails) or due to any of the changes in this PR. If it's not the fault of this PR I will go ahead and merge, as the code right now looks good.

PR #535, opened 2015-07-22
Widgets UX

sumitsahrawat:
+ [x] Demo `FlexBox` vs `Box`.
  - I'm still not completely sure what the difference is. (#536)
+ [x] Image widget: `cabal install HTTP` might need kernel restart.
  - Removed that line. Instead mentioned the fact in a markdown cell.
+ [x] Image widget: demo `#2` broken (%s/base64/encode64).
+ [x] Selection widgets cell `#1`.
+ [x] Selection widget plotting demo: Not in scope `unpack` and (`<$>`). Chart needed installed.
+ [x] Happy with selection and numeric examples. Simple but sensible.
+ [x] What happens when widgets do console input?
  - Explained and demonstrated.
+ [x] Numeric widgets duplicated in two places, same for selection widgets.
+ [x] Great use of marquee.
+ [x] Width changing code goes in the same place as the value-setting code.
+ [x] Parsec code, display widget once again in the end.

Further steps:

+ [x] Fix widget input issue by turning off stdin.
+ [x] Make clever use of Pattern Synonyms to remove `S` from singletons.


sumitsahrawat:
@gibiansky For ease of review, I split the last commit into two, one for the code changes, and one updating the examples.

We can now decide whether to go with TH or not.

gibiansky:
I like it. The changes are all boring and uneventful, which is good given the changes that were made. LGTM.

gibiansky:
Note that since this uses `PatternSynonyms`, this will not work with GHC 7.6. However I think that's fine given the upcoming release of 7.12 – soon 7.8, 7.10, and 7.12 will be the last three and the ones we will aim to support. It is fine for experimental features to only work on newer compilers.

PR #541, opened 2015-07-31
Fix compile error for ghc-7.10.2

qzchenwl:


sumitsahrawat:
@qzchenwl Thanks for the patch.

It seems like there is a type mismatch due to your edits. The build failure is not due to travis issues (unlike other build issues with IHaskell these days).

qzchenwl:
Travis uses ghc-7.10.1 which is a bit too old.

The type of [packageKeyPackageIdString] is `DynFlags -> PackageKey -> Maybe String` insteadof `DynFlags -> PackageKey -> String`.

[packageKeyPackageIdString]: https://github.com/ghc/ghc/blob/f9687caf337d409e4735d5bb4cf73a7dc629a58c/compiler/main/Packages.hs#L1450

sumitsahrawat:
Yup, I understand that. But, I don't think it'll be good for IHaskell to support a discontinuous range of GHC versions. After your patch, ihaskell supports GHC `6.3` to `7.10.2` excluding `7.10.1`.

I haven't played with `CPP` that much, but it should be possible to have that change apply only with GHC `7.10.2`. Am I right in saying this?

gibiansky:
Once 7.12 is released we can drop support for 7.10.1. For the time being we should continue to support 7.10.1, since many people have not yet upgraded to 7.10.2 – even the Stackage nightly isn't yet using 7.10.2. 

CPP is the right tool here – using `ghc_MIN_VERSION` macro, you should be able to fix this change to only apply for 7.10.2

qzchenwl:
You are right, I should fix it.

qzchenwl:
Fixed using `MIN_VERSION_ghc` macro.

sumitsahrawat:
@qzchenwl Take a look at [this](https://travis-ci.org/gibiansky/IHaskell/jobs/73807232) build log on `ghc-7.10.2`. Seems like you missed some of the issues, e.g. compare it with [this](https://travis-ci.org/gibiansky/IHaskell/jobs/73807231) build log using `ghc-7.10.1`.

qzchenwl:
@sumitsahrawat 

```
src/IHaskell/Eval/Evaluate.hs:199:66:
    Couldn't match expected type ‘[Char]’
                with actual type ‘Maybe String’
    In the second argument of ‘isPrefixOf’, namely ‘idString’
    In the first argument of ‘guard’, namely
      ‘(iHaskellPkgName `isPrefixOf` idString)’
src/IHaskell/Eval/Evaluate.hs:206:63:
    Couldn't match type ‘Maybe String’ with ‘[Char]’
    Expected type: [[Char]]
      Actual type: [Maybe String]
    In the second argument of ‘filter’, namely ‘packageNames’
    In the expression: filter (== iHaskellPkgName) packageNames
src/IHaskell/Eval/Evaluate.hs:214:88:
    Couldn't match expected type ‘[Char]’
                with actual type ‘Maybe String’
    In the second argument of ‘stripPrefix’, namely ‘pkgName’
    In the expression: stripPrefix initStr pkgName
src/IHaskell/Eval/Evaluate.hs:215:71:
    Couldn't match type ‘[Char]’ with ‘Maybe String’
    Expected type: [Maybe String]
      Actual type: [String]
    In the second argument of ‘notElem’, namely ‘broken’
    In the expression: pkgName `notElem` broken
src/IHaskell/Eval/Evaluate.hs:232:41:
    Couldn't match type ‘Maybe String’ with ‘[Char]’
    Expected type: [String]
      Actual type: [Maybe String]
    In the second argument of ‘map’, namely ‘displayPackages’
    In the expression: map toImportStmt displayPackages
```

Those are what I see before I fix ghc-7.10.2 compile error. Does the build include this patch?

sumitsahrawat:
It seems like the issue has been fixed. Nothing to worry about now :+1:

PR #544, opened 2015-08-03
Fix build issues

sumitsahrawat:
+ Remove ihaskell-widgets from build job for ghc-7.10.*
+ Add no-widgets option to build.sh


sumitsahrawat:
@gibiansky `hindent` gave an error in `Evaluate.hs`, which was fixable by replacing `#elif` with `#elseif`.

PR #545, opened 2015-08-04
Initial version of `interactive` for widgets.

sumitsahrawat:
Works only for single arguments (`Integer` or `Text`).

Example usage:

image

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -73,7 +73,7 @@ INSTALL_DIRS=`echo $INSTALLS | tr ' ' '\n' | sed 's#^#./#' | tr ' ' '\n'`
 echo CMD: cabal install --constraint "arithmoi -llvm" -j $INSTALL_DIRS --force-reinstalls --max-backjumps=-1 --reorder-goals
 cabal install --constraint "arithmoi -llvm" -j $INSTALL_DIRS --force-reinstalls --max-backjumps=-1 --reorder-goals
 
-if [ ! $2 = "no-widgets" ] && { [ $1 = "display" ] || [ $1 = "all" ]; } then
+if [ $2 != "no-widgets" ] && { [ $1 = "display" ] || [ $1 = "all" ]; } then
     cabal install ihaskell-display/ihaskell-widgets
 fi
 
--- a/ihaskell-display/ihaskell-widgets/ihaskell-widgets.cabal
+++ b/ihaskell-display/ihaskell-widgets/ihaskell-widgets.cabal
@@ -52,7 +52,8 @@ cabal-version:       >=1.10
 library
   -- Modules exported by the library.
   exposed-modules:     IHaskell.Display.Widgets
-  
+                       IHaskell.Display.Widgets.Interactive
+
   -- Modules included in this library but not exported.
   other-modules:       IHaskell.Display.Widgets.Button
                        IHaskell.Display.Widgets.Box.Box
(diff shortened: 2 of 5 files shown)


gibiansky:
I like this approach, actually. The code is a little bit ugly but it's very clear what's going on and it is very straightforward and logical.

Can it be generalized to N arguments? What issues have you run into with the generalization?

sumitsahrawat:
I had a miniature breakthrough in the generalization, with this helper function:

```haskell
type family WithTypes (ts :: [*]) (r :: *) :: * where
  WithTypes '[] r = r
  WithTypes (x ': xs) r = (x -> WithTypes xs r)

-- | Convert a function to one accepting arguments in the form of an HList
usingHList :: WithTypes ts r -> HList ts -> r
usingHList f RNil = f
usingHList f (Identity x :& xs) = usingHList (f x) xs
```

sumitsahrawat:
@gibiansky I made some progress, currently it looks like this:

image

Could still use some automation, and type magic.

gibiansky:
This looks good.

Is there a way to do all of this with less type-hackery? It's going to be very hard to accomodate things such as variable start/stop/step in the sliders with this set up, I think.

Perhaps something like this:
```haskell

type Start = Int
type Stop = Int
type Step = Int
type Label = String

data UI = IntUI Start Stop Step | StrUI Label
data Val = IntVal Int | StrVal String

interact :: [UI] -> ([Val] -> IO DisplayData) -> IO ()
interact = ...
```

This is less type-safe and requires a little bit more boilerplate, but the usability might be much higher.

Consider this:

```haskell
interact [IntUI 0 10 2, StrUI "My Label"] $ \[IntVal x, StrVal label] -> do
  let str = label ++ ": " ++ show x
  return $ plain str
```

The above is very flexible and fairly easy to understand, even though you don't get type checking between the `[UI]` and `[Val]` -- you can create invalid pattern matches very easily.

The point being: for this use case, ease of use trumps type safety. `interact` is meant only during interactive development, so having it be a little less safe is OK, because "compile time" and "run time" are actually one and the same (approximately). 

So, questions:

1. Is something like what I described possible?
2. Is it possible to do something like that with type safety without sacrificing usability and understandability?

Thoughts?

sumitsahrawat:
I had planned to add more type-hackery to the `FromWidget` typeclass. Currently, it associates to every type, a widget, a handler and a field. I was also going to add what it calculates to.

For example, an `Int` always calculates to a tuple `(Int, Int, Int)`. And, an `(Int, Int, Int)` evaluates to itself. Thus, we can write functions that accept either an `Int` or a tuple. This will be hard to implement, but more natural to use.

I'll try to make the current version work without having to explicitly using `usingHList` first. If it works (I think I'm close), then I'll try to make `interact` work with normal functions without requiring them to be list based. If not, then we can replace the clumsiness with runtime checks.

> Is something like what I described possible?

Yes, it's a good alternative in case I goof up with the types.

> Is it possible to do something like that with type safety without sacrificing usability and understandability?

We can use HList to do so, but that's similar to the current solution:

```haskell
interact :: Rec UI ts -> (HList ts -> IO DisplayData) -> IO ()
```

It'll also end up giving a not so nice syntax, so we have a lot of tradeoffs here. Also, those gigantic type signatures are no help to the users.

sumitsahrawat:
I always seem to miss the fact that interact can take more arguments than just a function, so I'll also have to make that work.

gibiansky:
@sumitsahrawat Let's try to get final code in soon?

sumitsahrawat:
@gibiansky I spent too much time trying to polish the code, so it's not as complete as I wanted it to be. I will get it done in a couple of days tops.

sumitsahrawat:
Better arguments :sparkles:

image

gibiansky:
Looks great. Ready to merge once it passes checks?

sumitsahrawat:
@gibiansky Yup. Strangely enough, I didn't see your comment before.

PR #557, opened 2015-08-21
Fix build script - round 2

sumitsahrawat:
The build script still didn't work as expected somehow. Hopefully this one will.


gibiansky:
It passed! Great!

PR #565, opened 2015-08-27
Removed IO, fixed JuicyPixels display

wyager:
JuicyPixels display wasn't working. I fixed it, removed any I/O, and changed it to render PNGs.

I had to remove a few image formats to get PNG support, but if anyone complains I can add conversion code.


wyager:
Further detail: the way this used to work was it would save the image to a JPEG file, load the bytes from the file, and then use those. Besides being circuitous, this method also broke because, at some point, `hGetContents` (by way of `prelude`'s `readFile`) started checking that it was only reading valid unicode or something, which meant that it would no longer read binary files like JPEGs without crashing.

gibiansky:
The build failure is due to the submitted code being misformatted. We enforce consistent code formatting by comparing with `hindent`. I'll go ahead and merge and fix this later.

PR #569, opened 2015-09-01
Make widget messages match with IPywidgets

sumitsahrawat:
The ihaskell-widgets should send messages identical to what the
IPywidgets send. This has not been tested as there are some unresolved
issues in IHaskell that surface only if ipywidgets is installed.

### DIFF
--- a/ihaskell-display/ihaskell-widgets/src/IHaskell/Display/Widgets/Bool/CheckBox.hs
+++ b/ihaskell-display/ihaskell-widgets/src/IHaskell/Display/Widgets/Bool/CheckBox.hs
@@ -40,11 +40,9 @@ mkCheckBox = do
   stateIO <- newIORef widgetState
 
   let widget = IPythonWidget uuid stateIO
-      initData = object
-                   ["model_name" .= str "WidgetModel", "widget_class" .= str "IPython.Checkbox"]
 
   -- Open a comm for this widget, and store it in the kernel state
-  widgetSendOpen widget initData $ toJSON widgetState
+  widgetSendOpen widget $ toJSON widgetState
 
   -- Return the image widget
   return widget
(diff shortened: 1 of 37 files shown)


sumitsahrawat:
@gibiansky It worked. I was able to create a button. I will test this more extensively today.

sumitsahrawat:
@gibiansky It still has some issues with other widgets. Buttons work just fine, but their appearance has changed a lot so it looks different.

gibiansky:
Is there anything I can do to help here? I'd like to get a 4.0-compatible release out soon.

sumitsahrawat:
It's been a week, so @jdfreder might be free to give us some guidance related to the new widget messaging spec.

Other than that, the only solution is to go for hit and trial methods, but we still wouldn't be sure how to deal with the bigger changes, such as the frontend trying to open a comm on kernel startup.

jdfreder:
The changes to the message spec involve sending the entire widget state along with the comm open message, this means the specially named `model_name` and `model_module` no longer are sent, since that information is already included in the widget state as `_model_name` and `_model_module`.  See difference in parsing here: https://github.com/jdfreder/ipywidgets/blob/98aff2cfbcabdbe2b08ff729f979c92c2b251ce5/ipywidgets/static/widgets/js/manager.js#L283-L300

The front-end now opens a comm on startup as the workaround included here: https://github.com/ipython/ipywidgets/pull/62

We wanted to get a change into the message spec for 4.0, but were instead told that wouldn't happen, and that we should instead implement the message as a comm message.  In the future, this will not be the case.

sumitsahrawat:
Thanks for the input @jdfreder :+1:

I understood the first part (sending initial state with `comm_open`), but it doesn't work for all widgets and causes a lot of `WrappedError`s in the console. I'll experiment with comparing messages and see if I can get it to work.

jdfreder:
Is it possible that you are missing some state, so the full state is not being sent to the front-end?  There's a lot of state that could be easy to miss, especially that in the base DOMWidget class: https://github.com/ipython/ipywidgets/blob/master/ipywidgets/widgets/widget.py#L455-L502

If that state isn't pushed, it may cause errors.

sumitsahrawat:
Yeah. It turns out that most of the example notebooks, which I'm using to test the current state, made use of boxes. Boxes have seen some twiddling with their properties, which would need to be fixed.

@jdfreder Thanks once again for the help :+1:

sumitsahrawat:
@gibiansky All the widgets we had have been brought up to what 4.0 requires.

Some new widgets have been added to `ipywidgets`, namely `Valid`, `Proxy` and `ProxyPlace`. I will add them along with example notebooks later.

PR #574, opened 2015-09-14
Adding new widgets

sumitsahrawat:
+ Valid: A widget to display boolean values.
+ Proxy: A widget that contains at most one DOMWidget in it.
+ PlaceProxy: A Proxy widget with a selector (string) value.

Will have to investigate how the selector is used by PlaceProxy.

### DIFF
--- a/ihaskell-display/ihaskell-widgets/ihaskell-widgets.cabal
+++ b/ihaskell-display/ihaskell-widgets/ihaskell-widgets.cabal
@@ -57,11 +57,14 @@ library
   -- Modules included in this library but not exported.
   other-modules:       IHaskell.Display.Widgets.Button
                        IHaskell.Display.Widgets.Box.Box
+                       IHaskell.Display.Widgets.Box.Proxy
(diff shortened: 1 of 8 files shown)


sumitsahrawat:
Note to self: The example notebooks will also need to be updated.

gibiansky:
Is the only thing remaining that is blocking https://github.com/jupyter/docker-demo-images/pull/71#issuecomment-139978343 making a release? If so, I can create a release and get it on Hackage / Stackage today(ish).

sumitsahrawat:
Yup, a release will set everything right. Currently the examples are not part of the `extra-source-files`, so we can make a release as it is.

I will work on the examples, and then we can have another release including new examples.

rgbkrk:
Yep, that's the only blocker.

gibiansky:
Updated versions on hackage should be available. `0.8.0.0`, `0.8.1.0`, `0.2.0.0` for `ihaskell` `ipython-kernel`, and `ihaskell-widgets`, respectively.

PR #586, opened 2015-10-09
Added some directives about using stack to manually install IHaskell,…

guillaumecherel:
… and how to deal with the libiconv issue in OS X.


sumitsahrawat:
LGTM. Will wait for @gibiansky to give it a go.

Thanks for the patch :+1:

PR #572, opened 2015-09-03
New message types for ipython-kernel

thomasjm:
Background: I've been working on something that's essentially a new Jupyter frontend in Haskell, and I needed several more IPython message types to be supported in `ipython-kernel`. This is kind of an ongoing process, but I thought I'd get a PR out with the current changes.

PR #600, opened 2015-10-28
Build failure fix

sumitsahrawat:
Seems like #572 had some errors hidden in the build logs.


thomasjm:
Oops, sorry about that @sumitsahrawat / thanks for cleaning up after me!

PR #614, opened 2015-12-06
Improve error handling in Mac OS X installer

kputnam:
My first few attempts to install via `macos-install.sh` failed at some steps. Subsequent commands failed because preceding steps failed and the important errors were buried in lots of console output.

This patch ensures the installer script halts on failure. I also added support for Python 3 installed via homebrew, which uses the name `pip3` instead of `pip`.

### DIFF
--- a/macos-install.sh
+++ b/macos-install.sh
@@ -1,4 +1,5 @@
 #!/bin/sh
+set -e
 
 # Installation for Mac OS X from IHaskell repo directory.
 # TODO Split out setup for installation from Hackage released versions.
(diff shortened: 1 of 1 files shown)


sumitsahrawat:
Great shell trickery. LGTM. Thanks for the hard work!

I don't have a mac, so will wait for @gibiansky to merge this.

kputnam:
Thanks for reviewing and merging! I really enjoy using IHaskell and only wish I knew about it earlier! :smile:

PR #630, opened 2016-01-24
Fix build faliures

sumitsahrawat:
+ Trying out new versions of happy and alex for ghc-7.6.*


gibiansky:
:+1:

We can stop supporting 7.6 in a few months!

PR #639, opened 2016-02-11
Support --version, -V in CLI

nicluo:
Sample Output:

    $ ihaskell --version
    0.8.3.0

    $ ihaskell -V
    0.8.3.0

Two points:

**First**: Adding <code>versionFlag</code> to <code>iHaskellArgs</code> displays the <code>-V --version</code> in the help command.
    
    IHaskell [COMMAND] ... [OPTIONS]
      Haskell for Interactive Computing.

      -? --help                  Display help message
      -V --version               Print version information

    IHaskell install [OPTIONS]
      Install the Jupyter kernelspec.

      -l --ghclib=<path>         Library directory for GHC.
         --debug                 Print debugging output from the kernel.
      -c --conf=<rc.hs>          File with commands to execute at start; replaces ~/.ihaskell/rc.hs.
         --prefix=<install-dir>  Installation prefix for kernelspec (see Jupyter's --prefix option)
      -? --help                  Display help message
         --stack                 Inherit environment from `stack` when it is installed

    IHaskell kernel [OPTIONS] <json-kernel-file>
      Invoke the IHaskell kernel.

      -l --ghclib=<path>         Library directory for GHC.
         --debug                 Print debugging output from the kernel.
      -c --conf=<rc.hs>          File with commands to execute at start; replaces ~/.ihaskell/rc.hs.
         --stack                 Inherit environment from `stack` when it is installed

    IHaskell convert [OPTIONS] <file>
      Convert between Literate Haskell (*.lhs) and Ipython notebooks (*.ipynb).

      -i --input=<file>          File to read.
      -o --output=<file>         File to write.
      -f --from=lhs|ipynb        Format of the file to read.
      -t --to=lhs|ipynb          Format of the file to write.
         --force                 Overwrite existing files with output.
      -s --style=bird|tex        Type of markup used for the literate haskell file
         --bird                  Literate haskell uses >
         --tex                   Literate haskell uses \begin{code}
      -? --help                  Display help message

Adding the subsequent lines to parseFlags allows the detection of <code>-V --version</code> flags so it can output the version accordingly.

    if "--version" `elem` flags || "-V" `elem` flags
      then Left VERSION_ipython_kernel

**Please help:** To me this seems like an incorrect place to put the check, but I'm not sure how to convert the version display into an <code>IHaskellMode</code>, if my intuition is correct.

**Second**: I removed the <code>if "--help" `elem` flags</code> in parseFlags because having them removed appears to not change its behavior in the command line.

### DIFF
--- a/src/IHaskell/Flags.hs
+++ b/src/IHaskell/Flags.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE CPP, NoImplicitPrelude, DeriveFunctor #-}
+{-# LANGUAGE NoImplicitPrelude, DeriveFunctor #-}
 
 module IHaskell.Flags (
     IHaskellMode(..),
@@ -58,7 +58,7 @@ data NotebookFormat = LhsMarkdown
   deriving (Eq, Show)
 
 -- Which mode IHaskell is being invoked in.
-data IHaskellMode = ShowHelp String
+data IHaskellMode = ShowDefault String
                   | InstallKernelSpec
                   | ConvertLhs
                   | Kernel (Maybe String)
@@ -71,9 +71,7 @@ parseFlags flags =
   in case modeIndex of
     Nothing ->
       -- Treat no mode as 'console'.
-      if "--version" `elem` flags || "-V" `elem` flags
-        then Left VERSION_ipython_kernel
-        else process ihaskellArgs flags
+      process ihaskellArgs flags
     Just 0 -> process ihaskellArgs flags
 
     Just idx ->
@@ -179,7 +177,7 @@ lhsStyleTex = LhsStyle "" "" "\\begin{code}" "\\end{code}" "\\begin{verbatim}" "
 ihaskellArgs :: Mode Args
 ihaskellArgs =
   let noMode = mode "IHaskell" defaultReport descr noArgs [helpFlag, versionFlag]
-      defaultReport = Args (ShowHelp helpStr) []
+      defaultReport = Args (ShowDefault helpStr) []
       descr = "Haskell for Interactive Computing."
       helpFlag = flagHelpSimple (add Help)
       versionFlag = flagVersion (add Version)
--- a/src/IHaskell/Flags.hs
+++ b/src/IHaskell/Flags.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE NoImplicitPrelude, DeriveFunctor #-}
+{-# LANGUAGE CPP, NoImplicitPrelude, DeriveFunctor #-}
 
 module IHaskell.Flags (
     IHaskellMode(..),
@@ -32,6 +32,7 @@ data Argument = ConfFile String     -- ^ A file with commands to load at startup
               | GhcLibDir String    -- ^ Where to find the GHC libraries.
               | KernelDebug         -- ^ Spew debugging output from the kernel.
               | Help                -- ^ Display help text.
+              | Version             -- ^ Display version text.
               | ConvertFrom String
               | ConvertTo String
               | ConvertFromFormat NotebookFormat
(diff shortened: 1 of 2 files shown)


nicluo:
@sumitsahrawat

src/IHaskell/Flags.hs
```diff
@@ -1,4 +1,4 @@
-{-# LANGUAGE NoImplicitPrelude, DeriveFunctor #-}
+{-# LANGUAGE CPP, NoImplicitPrelude, DeriveFunctor #-}
```
sumitsahrawat: Is this required? I can't see why it would be.

sumitsahrawat:
Thanks for the work @nicluo, it's greatly appreciated.

I was busy with college examinations, so I couldn't get back to you earlier. I'll take a look now.

src/IHaskell/Flags.hs
```diff
@@ -70,8 +71,8 @@ parseFlags flags =
   in case modeIndex of
     Nothing ->
       -- Treat no mode as 'console'.
-      if "--help" `elem` flags
-        then Left $ showText (Wrap 100) $ helpText [] HelpFormatAll ihaskellArgs
+      if "--version" `elem` flags || "-V" `elem` flags
+        then Left VERSION_ipython_kernel
```
nicluo: Removing <code>CPP</code> gives this error

    Building ihaskell-0.8.3.0...
    Preprocessing library ihaskell-0.8.3.0...
    [ 1 of 25] Compiling Paths_ihaskell   ( dist/build/autogen/Paths_ihaskell.hs, dist/build/Paths_ihaskell.o )
    [ 2 of 25] Compiling IHaskellPrelude  ( src/IHaskellPrelude.hs, dist/build/IHaskellPrelude.o )
    [ 3 of 25] Compiling IHaskell.Types   ( src/IHaskell/Types.hs, dist/build/IHaskell/Types.o )
    [ 4 of 25] Compiling StringUtils      ( src/StringUtils.hs, dist/build/StringUtils.o )
    [ 5 of 25] Compiling IHaskell.Eval.Util ( src/IHaskell/Eval/Util.hs, dist/build/IHaskell/Eval/Util.o )
    [ 6 of 25] Compiling IHaskell.Flags   ( src/IHaskell/Flags.hs, dist/build/IHaskell/Flags.o )

    src/IHaskell/Flags.hs:75:19:
        Not in scope: data constructor ‘VERSION_ipython_kernel’

sumitsahrawat: I'm not familiar with this particular usage of CPP. Seems like we'll need it.

sumitsahrawat: I investigated the `VERSION_` flags. They are defined by cabal for all packages. When you use `VERSION_ipython_kernel`, you're actually using the version for `ipython-kernel`, which might differ from the IHaskell version.

Looking at the old code, I think that the `if elem "--help" flags` bit was a hack to get help to show. The proper way of doing it would be to have the `parseFlags` function return a `Maybe Args` instead of `Either String Args`, and then working on the `Args` (if any) in the `ihaskell` function defined in `Main.hs`, as that is where all the flags are actually processed.

nicluo:
> you're actually using the version for ipython-kernel, which might differ from the IHaskell version.

The IHaskell version can be obtained in the following way

    import           Paths_ihaskell(version)
    import           Data.Version (showVersion)
    ...
    Usage: putStrLn (showVersion version)

I've replaced instanced of <code>VERSION_ipython_kernel</code> into the proper IHaskell version.

> working on the Args (if any) in the ihaskell function defined in Main.hs, as that is where all the flags are actually processed.

The showDefault (new commits) method now provides either Help or Version depending on the flags provided (help by default, version if explicitly <code>-V</code> or <code>--version</code>)

> Looking at the old code, I think that the if elem "--help" flags bit was a hack to get help to show. The proper way of doing it would be to have the parseFlags function return a Maybe Args instead of Either String Args

Without changing the parseFlags function, I was able to remove the <code>if elem "--help" flags</code> hack. I won't be modifying the parseFlags function further from now.

I think, with the latest commits, this PR should be ready to merge. <code>--version</code> is supported without hacking away at things.

Thanks a lot for the pointers @sumitsahrawat, and all the best to your examinations! :)

sumitsahrawat:
Yup, that looks stellar. I'll take a look at the build report, and merge if everything looks good :+1:

PR #640, opened 2016-02-16
README: drop lts versions

juhp:


sumitsahrawat:
Thanks for the fix :+1:

PR #652, opened 2016-03-25
ihaskell.cabal: add missing 'here' depend into testsuite

trofi:
Preprocessing test suite 'hspec' for ihaskell-0.8.4.0...

Hspec.hs:20:18:
    Could not find module ‘Data.String.Here’
    It is a member of the hidden package ‘here-1.2.7@here_1j09jYtlmBB7SYT9WO9k1d’.
    Perhaps you need to add ‘here’ to the build-depends in your .cabal file.
    Use -v to see a list of the files searched for.

Signed-off-by: Sergei Trofimovich <siarheit@google.com>


sumitsahrawat:
Thanks for the fix :+1:

Issue #656, opened 2016-04-14
caskroom capitalization 


aaronmck:
This is very minor, but when running the macos-install.sh script on my Mac, I had to change

brew tap caskroom/cask

to 

brew tap Caskroom/cask

in the XQuartz section to get the install to work.  I'm not sure if this is a recent brew change.

sumitsahrawat:
Thanks for the report. Would you be willing to send a quick patch?

aaronmck:
Sure, though I'm not sure if this affects other people.  It would be good if some other mac user could try a clean install and see if this is a unique problem.

Hodapp87:
What version of OS X are you on? I ran into issues installing on my Mac but I can't recall exactly what they were.

aaronmck:
10.11.  Again not sure if this affects others, though I found mention of a similar problem in other projects.

gibiansky:
Patched has been merged.

PR #657, opened 2016-04-14
adjust for changes in Caskroom

aaronmck:
as requested in issue #656, here's the patch.  It's unclear to me if this is a widespread issue or localized to specific setups.  Worth a clean install on a different machine to check.


sumitsahrawat:
I'm assuming that it won't increase issues, so here goes.

sumitsahrawat:
Thanks for taking the time :+1:

Issue #636, opened 2016-02-09
Recent Stack LTS versions don't seem to work


peddie:
I installed `ihaskell` completely from scratch using `stack` LTS 5.2 on two different machines running debian x86_64:

```
stack install --no-system-ghc ihaskell && stack exec --no-system-ghc ihaskell -- install --stack && stack exec --no-system-ghc jupyter -- notebook --debug --no-browser --ip=<IP address>
```

I encountered the same error in both cases:

```
ihaskell: /home/peddie/.stack/snapshots/x86_64-linux/lts-5.2/7.10.3/lib/x86_64-linux-ghc-7.10.3/cryptonite-0.10-9z0j8QI27Av2VIWw0mEkTO/libHScryptonite-0.10-9z0j8QI27Av2VIWw0mEkTO.a: unhandled ELF relocation(RelA) type 42

ihaskell: ihaskell: unable to load package `cryptonite-0.10'
[I 22:50:20.392 NotebookApp] KernelRestarter: restarting kernel (1/5)
```

`stack exec --no-system-ghc ghc-pkg -- list` has no trouble locating `cryptonite`.  `stack exec --no-system-ghc ghci` has no trouble loading and running functions from `Crypto.Random.EntropyPool` (to choose a random example).

My `kernel.json` file:

```
peddie@endeavour:~$ cat /home/peddie/.local/share/jupyter/kernels/haskell/kernel.json
{"display_name":"Haskell","argv":["/home/peddie/.stack/snapshots/x86_64-linux/lts-5.2/7.10.3/bin/ihaskell","kernel","{connection_file}","--ghclib","/home/peddie/.stack/programs/x86_64-linux/ghc-7.10.3/lib/ghc-7.10.3/","--stack"],"language":"haskell"}
```

This file was created with `/opt/ghc/7.10.3/lib/ghc-7.10.3/` in the `--ghclib` field of `kernel.json`, but the error did not go away when I manually replaced it with the path to `stack`'s libs.

EDIT: I forgot a critical piece of information: this issue does not occur when building with stackage LTS 5.1, which also uses `cryptonite-0.10`.

Jiehong:
I hot the same issue before, but all seems well with stack lts-5.6 so far.

peddie:
I still experience the exact same issue with stack lts-5.6 or 5.5 on debian sid x86_64, on multiple machines.

Hodapp87:
[...]

PR #660, opened 2016-05-10
Add a package flag to cryptonite

rpglover64:
Suggestion from: https://github.com/DanielG/ghc-mod/issues/762
Addresses: #636

I do not fully understand the implications of this change, but it fixed the problem for me (once I forced stack to rebuild `cryptonite`)


rpglover64:
Does not fully solve the problem; trying to render `diagrams` leads to an unhandled elf relocation in `libHSglib-0.13.2.2`.

Issue #665, opened 2016-05-28
In ipython-kernel, EasyKernel.replyTo creates incomplete KernelInfoReply


WolframKahl:
My first attempt at a custom kernel crashes with the following message:
```
[W 14:48:28.660 NotebookApp] 404 GET /nbextensions/widgets/notebook/js/extension.js?v=20160528144808 (::1) 1.56ms referer=http://localhost:8888/notebooks/PropLogic1.ipynb
KernelInfoRequest {header = MessageHeader {identifiers = ["\NUL\244\176\220Q"], parentHeader = Nothing, metadata = fromList [], messageId = UUID {uuidToString = "bb214027-fa68-4dbb-80ed-e0948a7b8156"}, sessionId = UUID {uuidToString = "76434A610DBA44D38763C4952471163C"}, username = "kahl", msgType = KernelInfoRequestMessage}}
ICalcCheck: src/IHaskell/IPython/EasyKernel.hs:(160,5)-(165,7): Missing field in record construction protocolVersion
````

Trying to hunt this down, I find that the definition of ```IHaskell.IPython.EasyKernel.replyTo``` contains:
```
replyTo config _ _ KernelInfoRequest{} replyHeader =
  return
    KernelInfoReply
      { header = replyHeader
      , languageInfo = kernelLanguageInfo config
      , implementation = "ipython-kernel.EasyKernel"
      , implementationVersion = "0.0"
      }
```
But in ```IHaskell.IPython.Types```, the relevant part of ```data Message``` is:
```
             -- | A response to a KernelInfoRequest.
               KernelInfoReply
                 { header :: MessageHeader
                 , protocolVersion :: String -- ^ current protocol version, major and minor
                 , banner :: String -- ^ Kernel information description e.g. (IHaskell 0.8.3.0 GHC 7.10.2)
                 , implementation :: String -- ^ e.g. IHaskell
                 , implementationVersion :: String -- ^ The version of the implementation
                 , languageInfo :: LanguageInfo
                 }
```
Now ```banner``` looks like something that should probably be populated from the ```KernelConfig``` or ```LanguageInfo```, but, being new to IPython/Jupyter, I have no idea what ```protocolVersion``` should be.

sumitsahrawat:
`protocolVersion` specifies the version of the messaging specification being used by the kernel. I think IHaskell uses the value `"5.0"` for it,

`banner` is a string shown as a one line description of the kernel in some places.

WolframKahl:
[...]

PR #666, opened 2016-05-29
Fix issue #665

sumitsahrawat:
Does this look right? @WolframKahl

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,6 +72,9 @@ script:
             cd "$TOP";
         done
 
+        # Ensure that ipython-kernel examples build successfully.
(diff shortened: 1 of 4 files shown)


gibiansky:
Does this need corresponding changes in the EasyKernel examples? Would be nice to get this merged just to avoid having a lingering PR

sumitsahrawat:
@gibiansky Yup, I was waiting for you to approve, but now that I think about it, the examples will need modifications.

Also, it would be good to have the examples covered by the travis builds. I'll see what I can do.

gibiansky:
Good point - the build should cover this.

PR #674, opened 2016-06-08
Travis quick fix: Travis fixes 3

sumitsahrawat:


gibiansky:
will incorporate in next travis test once the current set of builds finishes.

i am hoping that the stack caching will make builds way way faster

sumitsahrawat:
The way I see it, `stack` won't have to install any dependencies that don't change between consecutive builds. That should make everything go very fast.

gibiansky:
Yeah, that's the hope. This will take a few more slow rounds of iteration though, because caching won't work until the builds start passing, I think.

PR #697, opened 2016-10-04
Add CommInfoRequest and CommInfoReply

### LINKED ISSUE #694: Irrefutable pattern failed with IHaskell on Sage Math Cloud
I'm trying to get the IHaskell kernel on Sage Math Cloud, and I have it working when I open a notebook with another kernel, and then switch to IHaskell as the kernel afterwards. When opening a notebook using IHaskell, however, the kernel dies immediately. The error I'm getting is

```
ihaskell: src/IHaskell/IPython/Message/Parser.hs:(63,5)-(68,51): Irrefutable pattern failed for pattern Data.Aeson.Types.Internal.Success (messageType,
                                                                                         username,
                                                                                         messageUUID,
                                                                                         sessionUUID)
```

so I had a look on the issue tracker and found #532 which suggested running `ihaskell install --debug` and looking at the output. I'm attaching [jupyter-notebook.log](https://github.com/gibiansky/IHaskell/files/503561/jupyter-notebook.log.txt) but I believe the JSON that is causing the pattern match to fail is

```
{"username":"username","version":"5.0","msg_id":"0146B99A12FF46B18DE371FC7156C20D","msg_type":"comm_info_request","session":"9FCE21090B0C4690A60A3CF87A23871E"}
```

I noticed in #532 that IHaskell has certain requirements for the values of the required keys, so I was wondering if some incompatibility here is what's causing the pattern match to fail and the kernel to die? Or do you think the issue lies elsewhere?

emhoracek:
To fix #694.

This adds support for [`comm_info_request`](http://jupyter-client.readthedocs.io/en/latest/messaging.html#comm-info). I added `CommInfoRequest` and `CommInfoReply` Messages and all the Aeson bits for both. The `replyTo` functions are faked out for now with empty maps.

@vaibhavsagar says what's currently here is enough to fix his issue, but it would be better to actually respond with a list of open comms. :) I plan on doing that next.

### DIFF
--- a/ipython-kernel/src/IHaskell/IPython/EasyKernel.hs
+++ b/ipython-kernel/src/IHaskell/IPython/EasyKernel.hs
@@ -178,7 +178,7 @@ replyTo config _ _ CommInfoRequest{} replyHeader =
   return
     CommInfoReply
       { header = replyHeader
-      , commInfo = [] }
+      , commInfo = mempty }
 
 replyTo config _ interface ShutdownRequest { restartPending = pending } replyHeader = do
   liftIO $ writeChan (shellReplyChannel interface) $ ShutdownReply replyHeader pending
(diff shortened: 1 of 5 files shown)


emhoracek:
I added a real response for `comm_info_request`, so it's ready for review.

sumitsahrawat:
I'll let this idle for a couple of days so that @gibiansky can also review it if possible.

sumitsahrawat:
Thanks for the work :+1: @emhoracek and @vaibhavsagar

PR #705, opened 2016-12-04
macos-install.sh: remove install brew-cask

ickc:
As of [Release v0.60.0 · caskroom/homebrew-cask](https://github.com/caskroom/homebrew-cask/releases/tag/v0.60.0):

> it is no longer necessary to `brew install brew-cask`

Doing so will in fact return an error.


ickc:
Resolve #704 partially

PR #712, opened 2017-03-13
Switch back to parseUrl

erikd:


erikd:
This build failure had nothing to do with my change.

leftaroundabout:
[`parseUrlThrow`](http://hackage.haskell.org/package/http-client-0.5.6.1/docs/Network-HTTP-Client-Internal.html#v:parseUrlThrow) is fine, but it only works with `http-client>=0.4.30`. I adress this is in b52d484fd112048cf101f34af930d5017c825684

erikd:
@leftaroundabout Yes, that fixes it when building with cabal, but not when building with stack and the current stack resolver.

vaibhavsagar:
Is this still an issue?

erikd:
It was never an issue for me because I don't use stack.

vaibhavsagar:
Okay, closing.

Issue #700, opened 2016-10-14
There should be a documented method for setting RTS options


leftaroundabout:
Performance, memory comsuption etc. in Haskell often require particular options for the runtime system. With compiled programs as well as GHCi this can just be done from the command-line, like

```
ghci +RTS -M2g -N4
```

But `ihaskell` is generally not invoked from the command line but automatically by the `jupyter` process.

How can you set specific RTS options for the Haskell runtime – ideally on a per-notebook basis, alternatively just globally or for an entire `jupyter` process?

leftaroundabout:
Fixed in #714.

Thanks everybody!

PR #714, opened 2017-03-14
Enable RTS options

leftaroundabout:
Cf. https://github.com/gibiansky/IHaskell/issues/700

Run-time options are very useful to get the most out of GHC Haskell. I generally use them to actually get multithreading – which, after all, is often lauded as one of Haskell's big strengths – but also to limit rogue memory consumption, which is rather one of Haskell's weaknesses.

To enable these (apart from setting `-rtsopts` in the build) I included flags that write according fields in the kernelspec.

I'm not sure if my defaults are good, see discussion in 4c1b059.


leftaroundabout:
The [build failure](https://travis-ci.org/gibiansky/IHaskell/jobs/211057660) appears to be due to a broke `simplegeneric` Python package.

src/IHaskell/Flags.hs
```diff
@@ -95,6 +97,20 @@ help mode = showText (Wrap 100) $ helpText [] HelpFormatAll $ chooseMode mode
...
 ghcLibFlag = flagReq ["ghclib", "l"] (store GhcLibDir) "<path>" "Library directory for GHC."
 
+ghcRTSFlag :: Flag Args
+ghcRTSFlag = flagReq ["use-rtsopts"] storeRTS "\"<flags>\""
+                  "Runtime options (multithreading etc.). See `ghc +RTS -?`."
+ where storeRTS allRTSFlags (Args mode prev)
+          = fmap (Args mode . (:prev) . RTSFlags)
+              . parseRTS . words $ filter (/='"') allRTSFlags
+       parseRTS ("+RTS":fs)  -- Ignore if this is included (we already wrap
+           = parseRTS fs     -- the ihaskell-kernel call in +RTS <flags> -RTS anyway)
```
leftaroundabout: I suppose people might think the syntax is
```
ihaskell install --use-rtsopts="+RTS -N4 -RTS"
```
...but actually just
```
ihaskell install --use-rtsopts="-N4"
```
is needed. Rather than documenting this, it's probably easier to just strip the superfluous `±RTS` flags...

src/IHaskell/IPython.hs
```diff
@@ -53,6 +54,7 @@ data KernelSpecOptions =
 defaultKernelSpecOptions :: KernelSpecOptions
 defaultKernelSpecOptions = KernelSpecOptions
   { kernelSpecGhcLibdir = GHC.Paths.libdir
+  , kernelSpecRTSOptions = []
```
leftaroundabout: Default to _no_ RTS options for the time being (i.e. no memory limit, single processor, GC defaults from GHC.

In 4c1b0597e0381dbd6742543f32e230b2771f4def I propose better (?) defaults.

leftaroundabout:
Could somebody say something about this?

For me at least, the RTS options make IHaskell a lot more useful. Even without any defaults, it would clearly be sensible to have at least a simple way for everybody to set them up if needed, as given by the first two commits.

vaibhavsagar:
I have no strong feelings about this change one way or another. I'll wait for @gibiansky or @sumitsahrawat to comment.

leftaroundabout:
I was wondering why my IHaskell notebooks run so sluggish lately... turns out I forgot to rebase the RTS options stuff after the last pull from here.

Would somebody please care to merge this PR, or else explain briefly why not? @gibiansky ?

gibiansky:
@leftaroundabout I don't do much IHaskell maintenance lately, but this PR looks relatively sane, clear, and well-written. Provided that you've tested it, I have no problem w/ it getting merged.

@vaibhavsagar Given that I haven't been doing much here lately, I'll wait for you to go ahead and click merge? 

(If no action occurs in 5 days, I will merge it.)

sumitsahrawat:
Thanks for the work 👍

sumitsahrawat:
My bad. I didn't see you were here @gibiansky

gibiansky:
@sumitsahrawat nope, looks good

Issue #690, opened 2016-09-15
Support using IHaskell with GHC 8


vaibhavsagar:
Now that LTS-7.0 is out, I'd like to use IHaskell with GHC 8. It looks like this will require changes to at least `ghc-parser`. Is this something you're planning to support soon?

gibiansky:
This is perhaps a somewhat unsatisfactory answer, but I am currently working on a largely refactored and cleaned up IHaskell (`1.0`) that will support 7.10 and 8.0 (but not anything before 7.10). `ghc-parser` will no longer exist at that point, as it currently exists solely for the sake of maintaining support for pre-7.10 GHC versions. It is unlikely that IHaskell 1.0 will be out in the next few weeks, but probably within the next 2 months.

vaibhavsagar:
This is really exciting and not unsatisfactory at all, thanks for getting back to me! Feel free to close.

dabro:
I agree, great news.  For a stack and haskell newbie, how to run IHaskell in the meantime?  In a way that can be run in Jupyter from any directory?

vaibhavsagar:
Hi @dabro, you should be able to

1. `stack install ihaskell --resolver lts-6.17`
2. `ihaskell install --stack`
3. `stack exec jupyter notebook`

to get an IHaskell notebook from any directory.

dabro:
Thanks.  First line:
```
$ stack install ihaskell --resolver lts-6.17
Run from outside a project, using implicit global project config
Using resolver: lts-6.17 specified on command line
Downloaded lts-6.17 build plan.
No compiler found, expected minor version match with ghc-7.10.3 (x86_64) (based on resolver setting in /Users/browneej/.stack/global-project/stack.yaml).
Try running "stack setup" to install the correct GHC into /Users/browneej/.stack/programs/x86_64-osx/
```
I realize I'm missing a really simple stack concept--what is it?

dabro:
Follow-up on the above.

When I start jupyter notebook with `stack exec jupyter notebook --resolver lts-6.17`, I am able to run ihaskell notebooks within my IHaskell directory with no issues.

However, if I move them out to another data/project folder (same jupyter instance still running), the kernel crashes on launch:
```
[I 10:42:31.016 NotebookApp] Kernel started: f5049300-4408-4a51-bb2a-a1cdedc1a85f
Run from outside a project, using implicit global project config
[...]

PR #716, opened 2017-03-30
Support GHC 8

hyiltiz:
Solves #690 .


abarbu:
Thanks for submitting the PR to merge these but I'm not sure these changes belong here. I suspect I broke support for old GHCs and certainly broke support for old versions of jupyter/ipython and ipython-widgets. Backward compatibility is not something I need or care about.

ccressent:
@hyiltiz 
@vaibhavsagar

What's the plan regarding ghc support? Which versions should we aim to support? @abarbu said he only cares about ghc 8. Would supporting both 7.10 and 8 make sense?

I had started to fix the CI and update to ghc 8 in my own fork but it's clearly duplicating work done in this PR...
 
@gibiansky: are there commits we can salvage off your 1.0 branch? I've seen your message on the Haskell mailing list and I'm willing to help!

gibiansky:
It was, indeed, a fairly ambitious refactoring that never got completed. However, the goal of it was actually to drop GHC 7 support entirely, and have future versions support only GHC 8. So I don't mind that happening, actually, as long as there is a clear tag for the last version to support GHC 7 and info on the readme about that.

I don't mind having a fork around if that seems to be the best solution, but if someone is interested in avoiding a fork (which I think is OK but a bit confusing for users), and wants to put in the effort to merge the changes that make GHC 8 work (meanwhile removing support for GHC 7, if they so desire), I am happy to look at a PR and grant them commit access.

abarbu:
I agree that it's confusing to have a fork for GHC 8. Also, I don't want to maintain the fork, I just need ihaskell so I'm doing it :)

I think that backward-compatibility is a noble goal but is it the best use of the limited resources of this project? There are lots of other changes that would be more helpful for the vast majority of users.

Don't underestimate the huge amounts of extra overhead backward compatibility adds in terms of support, code, complexity, knowledge required to make changes, the inability to take advantage of new features easily, the complexity of the build system and the interpreter.  This will only get worse with time (with say.. backpack which users will want support for).

The older version of ihaskell will always live on in stack.

It's also not as easy as just targeting GHC 7, it means being compatible with old versions of every widget/display library that might have broken support for GHC 7. Also, folks that run older versions of GHC will probably also want to run older versions of ipython. jupyter has changed quite a bit, even the data model of widgets has (my branch only does enough to make things just about work but it breaks old versions in the process). Backward compatibility there is going to be another significant resource drain and isn't easy.

gibiansky:
@abarbu We're on the same page, I think, right?

I'm happy to drop support for GHC 7, as long as we just make a tag on Github for the last version that did support it and put a note in the README about that.

abarbu:
@gibiansky  Yup! I just wanted to warn away anyone that thinks backward compatibility isn't a big deal.

ccressent:
@abarbu raises a good point regarding limited resources and it looks like the consensus is to go with ghc 8 support only. That's perfectly fine and reasonable to me!

This PR is getting fairly large, I suggest the following course of action:
1. Finish the ghc 8 work and get this PR merged in
2. Fix and optimize the CI
3. Then we can work on open issues

@abarbu, @vaibhavsagar: what can I do to help with 1. and 2.?

vaibhavsagar:
I don't think anything further needs to be added to this PR. We should be able to merge it in as is and go from there.

gibiansky:
OK, in the interest of having something happen – @abarbu, I am happy to grant you commit access if you are willing to make a GHC 7 tag based on current master, merge the PR, and update the docs to reflect that IHaskell now only supports GHC 8. Do you have the time / desire to make that happen?

vaibhavsagar:
I'm more than happy to do this if @abarbu is not available!

abarbu:
Sure. I'm happy to make it happen when I get a chance in the next day or two if you give me access.

gibiansky:
@abarbu @vaibhavsagar You are both now listed as collaborators, and should be able to push to the repo and merge PRs. Please use this power wisely, and I appreciate your help with making this PR happen :)

vaibhavsagar:
I've created a tag and am about to merge this PR.

vaibhavsagar:
I've updated the docs to reflect that only GHC 8 is supported.

leftaroundabout:
@abarbu I agree that we shouldn't waste much time on backwards compatibility, but as of now I don't see your fears confirmed: the HEAD version if IHaskell is easy to get running with GHC-7.10, as per #747. Whilst this is so simple, I think we should keep that compatibility.

When we do abandon GHC-7, it should be made _really clear_ – the `base` dependency bumped and all `#if MIN_VERSION_ghc(8,0,0)` removed.

PR #719, opened 2017-06-04
Fix the tests

vaibhavsagar:
Thanks for raising this @leftaroundabout


vaibhavsagar:
\o/

vaibhavsagar:
I think this only works on `jupyter-client>=5.0` as mentioned [here](http://jupyter-client.readthedocs.io/en/latest/messaging.html#code-inputs).

PR #726, opened 2017-07-02
Add nix derivations

vaibhavsagar:
I'd like to be able to declaratively specify IHaskell as part of a dev environment using Nix.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ cabal.sandbox.config
 .tmp2
 .tmp3
 .stack-work
+result
--- a/README.md
+++ b/README.md
@@ -65,6 +65,19 @@ stack exec ihaskell -- install --stack
 stack exec jupyter -- notebook
 ```
 
+# Nix
+
+If you have the `nix` package manager installed, you can create an IHaskell
+notebook environment with one command. For example:
+
+```bash
+$ nix-build release.nix --arg packages "haskellPackages: [ haskellPackages.lens ]"
+<result path>
+$ <result path>/bin/ihaskell-notebook
+```
+
+It might take a while the first time, but subsequent builds will be much faster.
+
 ## Where are my packages?
 
 Stack manages separate environments for every package. By default your notebooks
--- a/release.nix
+++ b/release.nix
@@ -1,0 +1,69 @@
+{ packages ? (_: []), systemPackages ? (_: []) }:
+
+let
+  pkgs = import <nixpkgs> {};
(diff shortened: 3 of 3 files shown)


TikhonJelvis:
Do you have an example of how somebody would use this from Nix?

vaibhavsagar:
Currently this PR allows me to run

```bash
$ nix-build release.nix -A ihaskell
```

to generate an `ihaskell` executable that can then be installed as usual.

That isn't very useful on its own, but it's a good first step towards my eventual goal of being able to run a single command and bring up a notebook with additional dependencies specified as desired. I'm working on that [here](https://github.com/vaibhavsagar/experiments/tree/master/ihaskell-nix), and if you were to clone that repo, you would be able to do

```bash
$ nix-build ihaskell-nix/
<result path>
$ <result path>/ihaskell-notebook
```

to open an IHaskell notebook in the current directory.

vaibhavsagar:
A little force-pushing later, and you should be able to run

```bash
$ nix-build release.nix
<result path>
$ <result path>/bin/ihaskell-notebook
```

to get a notebook in the current directory with all `ihaskell-*` display modules loaded.

TikhonJelvis:
Perfect, thanks. 

It would be very useful to document that in `release.nix` somewhere.

vaibhavsagar:
I was thinking of putting it in the README after I got everything working. Current issues are:

- Defining a module in a notebook doesn't work, and complains about a missing `*.dyn_o`. No idea why this is, although I think it has something to do with the GHC compilation pipeline that IHaskell uses.
- The tests don't pass without my seemingly superfluous change. I think this has something to do with the `ghc-paths` module and how it interacts with the interpreter. Complicating things is the fact that doing a `cabal2nix --shell . > shell.nix` and running `cabal configure; cabal build; cabal install; cabal test` seems to work without any issues, as does `stack --nix build`.

I'd love advice/help with any of these things, because I'm relatively new to Haskell, IHaskell, and Nix :smile:

vaibhavsagar:
I decided that this is useful enough without trying to also fix the build and the missing `*.dyn_o` feature. I will track that in a separate PR.

wizzup:
@vaibhavsagar Is this working with ghc8? Do you have any plan to fix [NixOS #22047](https://github.com/NixOS/nixpkgs/issues/22047) ?

vaibhavsagar:
@wizzup This does work with GHC 8, and I'm hoping to bring it back in `nixpkgs` as soon as we get a new release on Hackage 😄

PR #729, opened 2017-07-19
Fix Nix build

vaibhavsagar:
Continuing from #726, I'd like to fix two issues with the Nix build:

- Defining a module in a notebook doesn't work, and complains about a missing `*.dyn_o`. No idea why this is, although I think it has something to do with the GHC compilation pipeline that IHaskell uses.
- The tests don't pass without my seemingly superfluous change. I think this has something to do with the `ghc-paths` module and how it interacts with the interpreter. Complicating things is the fact that doing a `cabal2nix --shell . > shell.nix` and running `cabal configure; cabal build; cabal install; cabal test` seems to work without any issues, as does `stack --nix build`.

I'd love advice/help with any of these things, because I'm relatively new to Haskell, IHaskell, and Nix :smile:

### DIFF
--- a/release.nix
+++ b/release.nix
@@ -20,6 +20,15 @@ let
         "ihaskell-widgets"
       ]);
   dontCheck = pkgs.haskell.lib.dontCheck;
+  fixPaths = package: pkgs.haskell.lib.overrideCabal package (old: {
+    preBuild = ''
+      export HOME=$TMP
+    '';
+    preCheck = ''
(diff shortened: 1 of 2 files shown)


vaibhavsagar:
I think I have a fix for the first issue in #742.

vaibhavsagar:
I no longer think broken tests when building with Nix are a dealbreaker. Closing this.

PR #735, opened 2017-07-31
Work with GHC 8.2

vaibhavsagar:
I'd like to support GHC 8.2 by the time the Stackage LTS package set including it is released. I've updated the package bounds and now I'm running into a GHC API change between <=8.0 and 8.2 where `extraPkgConfs` seems to have been removed and the order of the list it expects has reversed.

To check the current progress, run `nix-build release821.nix`.

### DIFF
--- a/ghc-parser/ghc-parser.cabal
+++ b/ghc-parser/ghc-parser.cabal
@@ -33,7 +33,7 @@ library
   -- other-modules:
   -- other-extensions:
   build-depends:       base                 >=4.6 && < 5,
-                       ghc                  >=7.6 && <8.1
+                       ghc                  >=7.6 && <8.3
 
   if   impl(ghc >= 7.6) && impl(ghc < 7.8)
     hs-source-dirs:      generic-src src-7.6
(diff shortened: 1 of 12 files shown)


vaibhavsagar:
Ready for testing! `nix-build release821.nix` should create a working notebook environment!

vaibhavsagar:
I am confused by the test failures and wondering what I changed to cause them.

vaibhavsagar:
I was hoping that upgrading to a new enough resolver would make the problem go away, but I no longer think it will and there's an issue in the codebase that needs to be addressed. I've added a small `failure.hs` that demonstrates the failure on GHC 8.2:

```
$ stack ghci failure.hs
*Main> eval "1 + 1"
*** Exception: Not in scope: ‘IHaskell.Display.displayFromChanEncoded’
No module named ‘IHaskell.Display’ is imported.
```

vaibhavsagar:
The issue appears to be that packages in GHC 8.0 and below are named e.g `ihaskell-1.2.3.4` but in 8.2 they're just `ihaskell`. I'm getting a different error now, so I'll call that progress :tada:

vaibhavsagar:
:tada: :fireworks: :champagne:

vaibhavsagar:
I'm going to split out `release821.nix` and the newer `stack.yaml` into a separate PR, and try to figure out how to test both 8.0 and 8.2 as part of the CI process.

Issue #728, opened 2017-07-15
IHaskell complains about missing "*.dyn_o" files when trying to use modules defined inline when built using Nix


vaibhavsagar:
When using a Nix-built IHaskell, the inline module definition feature does not work. Instead, this happens:
![ihaskell-dyn_o](https://user-images.githubusercontent.com/1525767/28238165-7948a0b6-6980-11e7-9c9f-cccbf9b524c2.png)

This doesn't happen when using a Stack-built IHaskell. In both cases, no `*.dyn_o` files are created.

vaibhavsagar:
I tried building IHaskell with `stack --nix`, and the resulting IHaskell works without issues.

vaibhavsagar:
It's worse than I thought: using a Nix-built IHaskell, loading source files with `:load` doesn't work either and results in the same error pictured above.

vaibhavsagar:
After asking around, it seems like what's happening is that GHC is using the `-dynamic` flag somewhere it shouldn't because of the way Nix wraps GHC. If we can track this down and selectively disable it this feature will work again.

PR #742, opened 2017-09-01
Enable dynamic loading

vaibhavsagar:
Addresses #728. Do you remember why you had this before @gibiansky?

### DIFF
--- a/release.nix
+++ b/release.nix
@@ -22,8 +22,15 @@ let
   dontCheck = pkgs.haskell.lib.dontCheck;
   haskellPackages = pkgs.haskellPackages.override {
     overrides = self: super: {
-      ihaskell       = dontCheck (
-                       self.callCabal2nix "ihaskell"          src                  {});
+      ihaskell       = pkgs.haskell.lib.overrideCabal (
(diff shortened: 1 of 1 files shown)


vaibhavsagar:
Looking at this further, it works by compiling to bytecode and doesn't create an `.o` file. I think there should be a better solution.

vaibhavsagar:
Getting IHaskell on Stack on Nix:

```bash
[me@nixos:~/repos/IHaskell]$ stack --nix build --fast
[me@nixos:~/repos/IHaskell]$ nix-shell -p 'python35.withPackages (ps: with ps; [ ipython ipykernel jupyter_client notebook ])'
[nix-shell:~/repos/IHaskell]$ .stack-work/install/x86_64-linux-nix/lts-9.2/8.0.2/bin/ihaskell install
[nix-shell:~/repos/IHaskell]$ export `stack --nix exec -- env | grep '^PATH='`:$PATH
[nix-shell:~/repos/IHaskell]$ jupyter notebook
```

vaibhavsagar:
Instead of changing the source, this latest change patches it only when `release.nix` is used. I'm happier with this as a stopgap solution.

PR #747, opened 2017-09-05
Restore compatibility with GHC-7 / old Cabal environments.

leftaroundabout:
Some of the changes done lately make the project compile _only_ with GHC-8, and only when there are no obsolete package-versions installed in the Cabal environment.

Here I adress these issues so IHaskell will also install on an old GHC-7.10 setup.

### DIFF
--- a/ihaskell.cabal
+++ b/ihaskell.cabal
@@ -68,7 +68,7 @@ library
                        ghc-paths            >=0.1,
                        haskeline            -any,
                        hlint                >=1.9,
-                       haskell-src-exts     >=1.16,
+                       haskell-src-exts     >=1.18,
                        http-client          >= 0.4,
                        http-client-tls      >= 0.2,
                        mtl                  >=2.1,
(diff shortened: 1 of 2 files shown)


vaibhavsagar:
Thanks for this PR! Dropping 7.10 compatibility is intentional as per https://github.com/gibiansky/IHaskell/pull/716#issuecomment-300482782 and the `README`. I was actually planning to remove more of the <8.0 stuff because it complicates the codebase and makes adding support for future GHCs harder.

Can you tell me why you want 7.10 support in `master` when you can use the `GHC7` tag?

leftaroundabout:
Well, actually I didn't realise there's a GHC-7 tag... _where_ does the README actually tell me about it?

The reason I want to stay in `master` is of course to easily keep up-to-date. Rather than switching to a legacy branch, I'd switch the compiler – I'm not overly concerned with long backwards compatibility, but I'm also not a fan of Stack's “everything in realtime” approach. GHC-7.10 isn't, like, Fortran 95 gfortran or something, it's no big deal to keep compatibility with it...

vaibhavsagar:
It's true that 7.10 is not that old, but I'm already running into issues with #735 because the GHC API has noticeably changed between 8.0 and 8.2. I don't see the benefit of keeping backwards compatibility when no new features have been added since the `GHC7` tag (other than GHC 8 support). I can update the README to mention the tag, would that be sufficient resolution?

leftaroundabout:
Ok, just the lack of support should then also be reflected in the Cabal dependencies. #748

PR #748, opened 2017-09-06
Properly un-support GHC-7 (and other old stuff)

leftaroundabout:
In #716, it was decided to not waste effort on supporting GHC-7.

This PR adjusts the dependencies to make this clear when Cabal-compiling the project with older versions.

(But see also #747, for how GHC-7.10 _could_ still be supported.)


vaibhavsagar:
Thanks!

PR #755, opened 2017-10-01
Nix improvements

### LINKED ISSUE #728: IHaskell complains about missing "*.dyn_o" files when trying to use modules defined inline when built using Nix

vaibhavsagar:
This fixes #728 and generally makes the Nix option more pleasant to use.


vaibhavsagar:
working-modules
compiled-output
Victory!

PR #757, opened 2017-10-08
Make silent and store_hitsory optional params

zainab-ali:
According to the [Jupyter messaging protocol](http://jupyter-client.readthedocs.io/en/latest/messaging.html#execute), `silent` and `store_history` should be optional parameters.  

`silent` should default to `False` while `store_history` should default to the opposite of `silent`.


vaibhavsagar:
Wonderful, thanks for this PR!

PR #759, opened 2017-10-15
Test with GHC 8.2

vaibhavsagar:
Following on from https://github.com/gibiansky/IHaskell/pull/735#issuecomment-336695440


vaibhavsagar:
I would like to clean this up and update e.g. package versions before merging this.

PR #769, opened 2017-10-21
release{,-8.2}.nix: run tests after build

vaibhavsagar:
This allows us to test Nix-built IHaskell on CI!


vaibhavsagar:
I'm allowing the Nix builds to fail for now because they take too long to build.

PR #780, opened 2017-11-06
Fix #779: add missing installation step to README

### LINKED ISSUE #779: README installation steps missing / out of order
I tried installing this on my Mac and it failed on the "ihaskell install --stack" step because I didn't have ipython or jupyter.

I also don't see how "pip3 install -r requirements.txt" makes sense before cloning the repo in the Linux case.

aostiles:


vaibhavsagar:
Thanks!

PR #798, opened 2017-12-14
Build 'jupyter console' wrappers with nix installation

unode:
For those of us that prefer the console version of jupyter.
This adds `ihaskell-console` which runs `jupyter console --kernel=haskell`.

On an unrelated note, any plans to make a release anytime in the near future?
I'd like to have ihaskell available in nixpkgs but the last release is a couple of years old.

### DIFF
--- a/release-8.2.nix
+++ b/release-8.2.nix
@@ -98,11 +98,11 @@ let
   };
   ihaskellEnv = haskellPackages.ghcWithPackages (self: [ self.ihaskell ] ++ packages self);
   jupyter = nixpkgs.python3.withPackages (ps: [ ps.jupyter ps.notebook ]);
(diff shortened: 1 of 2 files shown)


vaibhavsagar:
Thanks for this contribution! The most recent version of IHaskell in nixpkgs is only a few months old and corresponds to the latest release on Hackage. I don't have package uploading privileges so I was waiting for `lts-10` to come out before pinging @gibiansky to make a new release. Is there a reason you can't override the derivation in nixpkgs with `master` for the time being?

unode:
Oh I didn't notice the release on Hackage. I see 0.9.0.1 there. This will do fine.
I was referring to [github releases](https://github.com/gibiansky/IHaskell/releases) which only includes 0.8.3.0.

PR #817, opened 2018-02-06
Support GHC 8.4

vaibhavsagar:
It's in Nixpkgs `master` so it's possible to test against it.


vaibhavsagar:
It compiles and the tests pass!

PR #832, opened 2018-03-06
(Idiomatically) customise size of diagrams shown in notebooks

leftaroundabout:
1f852097f9cd373a31fc5d94b28ede12352dd401 exposed an `IORef` that could be modified in order to tweak the size at which diagrams are rendered for display. This works, but it seems very un-Haskellish and also isn't very flexible when other properties than the height should be fixed.

This PR takes an alternative approach: it offers a new type that wraps a size specification with a diagram. This way, it can be tweaked in a declarative way. Also, I've changed the default behaviour: previously, the _height_ would get fixed to 300. But this is problematic when displaying diagrams of different aspect ratios, in particular, diagrams with small height used to get stretched extremely wide.

Example of the new behaviour: http://nbviewer.jupyter.org/gist/leftaroundabout/ee800328d4cda035955878dd680e677b

### DIFF
--- a/ihaskell-display/ihaskell-diagrams/IHaskell/Display/Diagrams.hs
+++ b/ihaskell-display/ihaskell-diagrams/IHaskell/Display/Diagrams.hs
@@ -62,8 +62,12 @@ withSizeSpec spec renderable = ManuallySized renderable imgWidth imgHeight
          V2 (Just w) (Just h) -> (w, h)
          V2 (Just w) Nothing  -> (w, w/aspect)
          V2 Nothing (Just h)  -> (aspect*h, h)
-         V2 Nothing Nothing   -> (aspect*defaultHeight, defaultHeight)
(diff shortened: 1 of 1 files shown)


leftaroundabout:
This PR is not complete yet: the `ManuallySized` wrapper should also be used at least for `Animation`. Waiting for feedback before I start that.

vaibhavsagar:
I find the implementation more complex (and idiomatic) than my approach, but I like the results and I'm happy to merge if you continue down this path with `Animation`.

leftaroundabout:
Done; see again [the demo notebook](http://nbviewer.jupyter.org/gist/leftaroundabout/ee800328d4cda035955878dd680e677b).

vaibhavsagar:
Awesome, thanks! Merge at will.

(I changed my mind and went ahead and merged it, thanks again!)

PR #839, opened 2018-03-16
Test with GHC 8.4

vaibhavsagar:


vaibhavsagar:
I think it makes sense to merge this and gradually re-enable the other packages as dependencies become available.

PR #844, opened 2018-03-24
Support using 'stack --docker'

jcberentsen:
Tested on OSX

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,29 @@ Currently the component that takes the longest time to compile is
 `ihaskell-widgets`, so if you're in a hurry you may want to comment that out in
 `stack.yaml`.
 
+## Stack development with Docker
+This is an alternative way to use Docker than above, taking advantage of stack's Docker support.
+Makes it easy to install new packages with `stack --docker install <pkg>` without having to rebuild a docker image.
+Like the other Docker workflow, this doesn't require any host dependecies to be installed.
+
+```bash
+docker build -t ihaskell-dev docker
+stack --docker setup
+stack --docker install
+stack --docker exec ihaskell -- install --stack
+stack --docker exec jupyter -- notebook --ip=0.0.0.0 notebooks
+```
+
+Everything in the LTS can be made available!
+To add a package outside the LTS, simply add it to the `stack.yaml` file  (See: "Where are my packages?" below).
+Then install the package with stack before restarting `jupyter`
+
+```bash
+# after adding details about mypackage to stack.yaml
+stack --docker install mypackage
+stack --docker exec jupyter -- notebook notebooks
+```
+
 ## Nix
 
 If you have the `nix` package manager installed, you can create an IHaskell
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,13 @@
+FROM fpco/stack-build:lts-11.1
+# docker build -t ihaskell-dev docker
+
+# Install all necessary Ubuntu packages
+RUN apt-get update && apt-get install -y python3-pip libgmp-dev libmagic-dev libtinfo-dev libzmq3-dev libcairo2-dev libpango1.0-dev libblas-dev liblapack-dev gcc g++ && \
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -81,6 +81,29 @@ Currently the component that takes the longest time to compile is
...
 
+## Stack development with Docker
+This is an alternative way to use Docker than above, taking advantage of stack's Docker support.
+Makes it easy to install new packages with `stack --docker install <pkg>` without having to rebuild a docker image.
+Like the other Docker workflow, this doesn't require any host dependecies to be installed.
+
+```
+docker build -t ihaskell-dev docker
+stack --docker setup
+stacl --docker install
```
vaibhavsagar: Please fix this typo.

README.md
```diff
@@ -81,6 +81,29 @@ Currently the component that takes the longest time to compile is
...
+stack --docker exec ihaskell -- install --stack
+stack --docker exec jupyter -- notebook --ip=0.0.0.0 notebooks
+```
+
+Everything in the LTS should be available!
```
vaibhavsagar: I don't see how this is true. Does the `Dockerfile` somehow download and install all packages in the LTS? If not, please reword this to say something like "Everything in the LTS can be made available!"

README.md
```diff
@@ -81,6 +81,29 @@ Currently the component that takes the longest time to compile is
...
+To add a a package outside the LTS, simpy add it to the `stack.yaml` file  (See: "Where are my packages?" below).
```
vaibhavsagar: Following on from my comment above, please change this to "To add a package"

README.md
```diff
@@ -81,6 +81,29 @@ Currently the component that takes the longest time to compile is
...
+Then install the package with stack before restarting `jupyter`
+
+```
+# add details about mypackage to stack.yaml
```
vaibhavsagar: Can this be reworded to something like this?

```
# after adding details about mypackage to stack.yaml
```

README.md
```diff
@@ -81,6 +81,29 @@ Currently the component that takes the longest time to compile is
...
+stack --docker install mypackage
+stack --docker exec jupyter -- notebook notebooks
+```
+
```
vaibhavsagar: One newline instead of two, please.

README.md
```diff
@@ -81,6 +81,29 @@ Currently the component that takes the longest time to compile is
...
+This is an alternative way to use Docker than above, taking advantage of stack's Docker support.
+Makes it easy to install new packages with `stack --docker install <pkg>` without having to rebuild a docker image.
+Like the other Docker workflow, this doesn't require any host dependecies to be installed.
+
+```
+docker build -t ihaskell-dev docker
+stack --docker setup
+stacl --docker install
+stack --docker exec ihaskell -- install --stack
+stack --docker exec jupyter -- notebook --ip=0.0.0.0 notebooks
```
vaibhavsagar: Can we move some of these instructions out of the `README` and into the `Dockerfile`? I think everything up until the final `stack --docker exec jupyter -- notebook --ip=0.0.0.0 notebooks` should be safe to put in the `Dockerfile`. If not, why?

vaibhavsagar:
Thanks so much for this contribution! I'm generally happy with this and looking forward to merging it eventually, but I have a few concerns/questions first.

README.md
vaibhavsagar: Also s/simpy/simply/.

README.md
jcberentsen: Agree, "can be made available" is more accurate

README.md
jcberentsen: The docker image used doesn't actually contain the IHaskell source, so `stack` can't see the `stack.yaml` which is outside the docker build context, can it?

All the build artifacts and packages will be on the _host_ (volume mounted into the docker container), which allows installing additional packages without rebuilding the docker image and keeping the docker container ephemeral.

vaibhavsagar: You're right, I had forgotten about this.

vaibhavsagar:
Thanks again!

PR #848, opened 2018-03-31
Dockerfile: update to work with mybinder.org

vaibhavsagar:


vaibhavsagar:
No idea why this isn't working on mybinder.org, this works when I run it locally:

```bash
$ docker build -t ihaskell:latest .
<wait forever...>
$ docker run -it -p8888:8888 ihaskell:latest
```

Maybe it's too large (11.7GB)?

vaibhavsagar:
Update: it seems to work now! Merging this.

PR #855, opened 2018-04-25
src/IHaskell/Eval/Parser.hs: support lowercase 'language' pragmas

vaibhavsagar:


Wizek:
And I've also extended with this: https://github.com/gibiansky/IHaskell/pull/856 😉

GHC does support it, and even though IIRC not many people do this (could be interesting to grep hackage) some might, and this way issue may not come up in the future.

ekmett:
Note: pragmas are case insensitive, so you should use `toLower` rather than trying just the two extremes.

```haskell
    {-# Language FunctionalDependencies #-}
```

and

```haskell
    {-# LaNgUaGe CPP #-}
```

work just as well.

vaibhavsagar:
I didn't realise they were completely case-insensitive! Addressed in https://github.com/gibiansky/IHaskell/pull/857.

PR #856, opened 2018-04-25
Also support first-capital, as GHC does too.

Wizek:


vaibhavsagar:
Sorry, I should have merged this before rebasing.

Wizek:
It seems it did end up on master, so that's okay.

Issue #862, opened 2018-05-10
ihaskell-display modules not loaded with nix build


drever:
The ihaskell-display modules are not loaded when building with nix. E.g. when running the blaze demo from the demo notebook the following error occurs:

> <interactive>:1:1: error:
>    Failed to load interfaces for 'Text.Blaze.Html4.Strict'

The commit [release.nix: use extend instead of override](https://github.com/gibiansky/IHaskell/commit/6536096103488438ed62e23a0efa8819d1754200) might be related. Please let me know if I should provide any further information.

vaibhavsagar:
They're not loaded by default, you need to specify them as additional packages.

drever:
That was it. Thanks for the amazingly fast response and have a good flight!

PR #863, opened 2018-05-10
Update README.md

drever:
This is a follow up to issue #862 regarding the specification of extra IHaskell-modules.


vaibhavsagar:
Thanks!

Issue #865, opened 2018-05-27
Kernel takes on the order of 10 seconds to warm up


benzrf:
If I make a new notebook and evaluate something simple like `3` or `"a"`, it will take on the order of 10 seconds for the kernel to respond. Once it does, however, it seems to subsequently work fine.

I tried installing with profiling enabled and then manually compiling `Main.hs`, and I got [some profiling information](https://benzrf.com/uploads/cf00a56c9737e08a.prof) (warning: 4.6M file). The most relevant info:
```
                                                                                                               individual      inherited
COST CENTRE    MODULE                    SRC                                               no.      entries  %time %alloc   %time %alloc
  interpret    IHaskell.Eval.Evaluate    src/IHaskell/Eval/Evaluate.hs:(166,1)-(187,28)     3102         1   88.6    3.4    99.5   96.3
```

I'm running IHaskell via Nix by using the `ihaskell-notebook` command from a shell launched with `nix-shell -p 'ihaskell.override {inherit (haskell.packages.ghc802) ghcWithPackages;}'`. (Strictly, the profiling info I dumped was in a slightly different environment where I'd added one or two packages and had enabled profiling, but the issue is similar (or worse) in the listed environment.)

vaibhavsagar:
It's not surprising that a lot of time is spent in `interpret` when that is what actually ends up interpreting Haskell code. Maybe the [initialisation](https://github.com/gibiansky/IHaskell/blob/560ccbf15be2f711b40abf5e61400cbc7093f137/main/Main.hs#L175-L195) code is contributing to the slow startup time?

benzrf:
This profiling info is from *only* starting the kernel and then evaluating `3` and then shutting down the kernel, so it should accurately reflect the startup cost (I think!).

benzrf:
I guess I could try, but the profiling information does say that it was about 80% in `interpret` *non-inherited*, meaning that the time was being spent somewhere that isn't listed anywhere else.

vaibhavsagar:
Makes sense! I'm seeing more like 3 seconds on my machine though, can you try [this](https://github.com/vaibhavsagar/notebooks/tree/fe4685d90581e584a71e114df161f9f52cacab0f/tarjan) and see if you are still getting ~10 second startup times?

vaibhavsagar:
[...]

Issue #659, opened 2016-04-27
ihaskell: Error casting types in Evaluate.hs


vaibhavsagar:
Steps to reproduce:

1. `pipsi install jupyter-core` (and install other packages so that Jupyter works)
2. `git clone git@github.com:gibiansky/IHaskell.git; cd IHaskell; stack install`
3. `source ~/.local/venvs/jupyter-core/bin/activate`
4. `ihaskell install`
5. `stack exec jupyter -- notebook` or `stack exec jupyter -- console --kernel haskell`
6. Try using a notebook or the REPL.

This only started happening on Ubuntu 16.04. Could this be the issue?

bzdzb:
YMMV, but I was able to get IHaskell installed on Ubuntu 16.04 by doing:

1. `pipsi install jupyter-core` (and install other packages so that Jupyter works)
2. `apt-get install -y libmagic-dev libncurses5-dev pkg-config libcairo2-dev liblapack-dev libblas-dev libpangocairo-1.0 libpango1.0-dev libzmq5-dev libzmq5`
3. `git clone git@github.com:gibiansky/IHaskell.git && cd IHaskell`
4. `source ~/.local/venvs/jupyter-core/bin/activate`
5. `stack setup && stack build && stack install`
6. `pip install jupyter` (after jupyter-core venv is activated)
7. `ihaskell install`
8. `stack exec jupyter -- notebook` || `stack exec jupyter -- console --kernel haskell`

Just a guess, but I suspect the problem you're seeing may be related to a ghc version incompatibility with one of the dependency versions.

vaibhavsagar:
Thanks for the update! I followed your instructions and this time I'm getting #636 but with `cryptonite-0.15`. This is a step in the right direction :smile:

thomasjm:
I'm getting this same problem, and confused why @bzdzb's instructions helped (what does having Jupyter in a virtualenv have to do with IHaskell building properly?)

I'm building the HEAD version of IHaskell using the current resolver `nightly-2016-03-25` and having no luck. I put some print statements in the function in question:

```haskell
-- | Compile a string and extract a value from it. Effectively extract the result of an expression
-- from inside the notebook environment.
extractValue :: Typeable a => String -> Interpreter a
extractValue expr = do
  print "The expr:"
  print expr
  compiled <- dynCompileExpr expr
  print "The compiled thing: "
  print compiled
  case fromDynamic compiled of
[...]

Issue #655, opened 2016-04-13
kernel keeps dying


inktrap:
Maybe this is a problem caused by me, but ihaskell on debian testing does not work for me.

I installed it via cabal (see below) and registered the kernel. I get the following message that cryptonite is not installed (``ihaskell: ihaskell: unable to load package `cryptonite-0.8'``) but it is.

# steps

Here are the detailed steps that I did:

 - prerequisites: a working haskell requirement: ``sudo apt-get install haskell-platform ghc``
 - remove old config: ``rm -rf .config/ipython``
 - install ipython3: ``sudo pip3 install jupyter ipython jinja2 zmq tornado``
 - start the notebook to see if it works: ``jupyter notebook --ip=127.0.0.1``
 - install zeromq-3: ``sudo apt-get install libzmq3-dev``
 - if it works, install ihaskell via cabal: ``cabal install ihaskell --reorder-goals``
 - ``sudo ln -s $(which ipython3) /usr/local/bin/ipython``
 - clear any alias that you might have set for ipython or jupyter
 - add the new kernel to ipython: ``ihaskell install``
 - start the notebook to see if it works: ``jupyter notebook --ip=127.0.0.1``
 - status: haskell kernel keeps dying

That didn't work, so I unregistered all the packages and tried it via the install script from the repo (``ubuntu-install.sh``). However the result is the same and the kernel keeps dying.

 - removed all installations via pip and installed via git and scripts, same problem

# Errorlogs

## console

```
vh@box ~ [n] % jupyter console --debug --kernel haskell
[ZMQTerminalIPythonApp] Searching ['/home/vh', '/home/vh/.jupyter', '/usr/etc/jupyter', '/usr/local/etc/jupyter', '/etc/jupyter'] for config files
[ZMQTerminalIPythonApp] Looking for jupyter_config in /etc/jupyter
[ZMQTerminalIPythonApp] Looking for jupyter_config in /usr/local/etc/jupyter
[ZMQTerminalIPythonApp] Looking for jupyter_config in /usr/etc/jupyter
[ZMQTerminalIPythonApp] Looking for jupyter_config in /home/vh/.jupyter
[ZMQTerminalIPythonApp] Looking for jupyter_config in /home/vh
[ZMQTerminalIPythonApp] Looking for jupyter_console_config in /etc/jupyter
[ZMQTerminalIPythonApp] Looking for jupyter_console_config in /usr/local/etc/jupyter
[...]

PR #867, opened 2018-05-29
ipython-kernel: use cryptonite instead of SHA

vaibhavsagar:
This might make a difference to #865, and `cryptonite` is elsewhere in the dependency tree for IHaskell anyway!


vaibhavsagar:
Hmm, it seems I was mistaken about `cryptonite` (or at least I can't find where it is depended on). I know we've had issues with it in the past so I assumed it was a dependency: #636, #659, #655

vaibhavsagar:
Okay, it looks like `tls` is what depends on `cryptonite`.

PR #886, opened 2018-07-09
Jupyterlab compatibility

MMesch:
This branch is compatible with jupyterlab. Some problems remain, for example how to add the IHaskell codemirror mode to jupyterlab. Also the haskell specific renderer have to be added in a separate extension.


vaibhavsagar:
Can you fix the conflicts please :smile:? I'd like to run our test suite on these changes so I can confirm that they don't affect the existing Jupyter kernel support.

MMesch:
on my computer it runs, but I didn't manage to get it running on Binder with docker so far ...

MMesch:
I finally figured out how to add the ihaskell codemirror mode to jupyterlab in an extension. The help mode is still missing but won't be difficult to implement ... the notebook functionality is not affected as far as I can see it @vaibhavsagar

MMesch:
:h :t :k all seem to work ... from my side it runs smooth and it could be merged if you want it.

vaibhavsagar:
This is great, thanks so much! Would it be at all possible for you to split up the changes into smaller commits? For example you have the `status` and `pygments` changes together when I think they could be separate.

MMesch:
some remarks:

_initial idle message:_
it is possible that the initial idle message is not needed anymore at some point because jupyterlab is changing as well. On the other hand, the ipython kernel seems to send one and it probably doesn't hurt to keep sending it to be on the safe side.

_status message for all reply messages:_
these weren't required for jupyterlab but seem to be required by the jupyter messaging protocoll

_pygment lexer info:_
I think this is also not required for jupyterlab because codemirror does the syntaxx highlighting in the notebook. I think pygments is needed for nbconvert and taken as the language name in case that it is not set (correct me if I'm wrong). It seems reasonable to explicitely define it.

_remove 'source' property:_
this is required

_jupyter labextension_
I think this extension could be grown to provide some haskell-specific mime types (hoogle display?), keyboard shortcuts etc...

vaibhavsagar:
Perfect, this is exactly what I was hoping for. Thanks again!

PR #890, opened 2018-07-18
use a colorful haskell logo

MMesch:
A colorful logo looks much better in the Kernel gallery (see screenshot) I tested this with jupyterlab and changed the file to an .svg whereever I could find it.

screenshot from 2018-07-18 16-20-52


vaibhavsagar:
Wonderful, thanks for doing this!

PR #891, opened 2018-07-19
updated version numbers

MMesch:
the IHaskell package (uses new mime types) depends on the new ipython-kernel (defines the Mime types). The version numbers need to be adopted. I think this should be it, but can you check it @vaibhavsagar ?


vaibhavsagar:
I would prefer `0.9.1.0` for `ipython-kernel` and `0.9.1.0` for `ihaskell`. I was planning to cut a new release after implementing GHC 8.6 support but if you want to use your new changes from Hackage I can put the new release up in the next day or two :smiley:.

MMesch:
>  if you want to use your new changes from Hackage I can put the new release up in the next day or two

that would be great ;)

PR #893, opened 2018-07-24
updated package.json for npmjs.org upload

MMesch:
The ihaskell extension can now be installed from nomjs.org directly.


vaibhavsagar:
Thanks!

PR #902, opened 2018-08-11
Support GHC 8.6

vaibhavsagar:


vaibhavsagar:
I added enough overrides to build the tests for `ihaskell` (uncomment `doCheck = true;` in `release-8.6.nix`) but they seem to spin indefinitely without producing useful output. Not sure where to go from here.

vaibhavsagar:
The tests pass! The issue is that the `Eval` tests are agonisingly slow and Jupyter times out waiting for the kernel to start without `--enable-executable-dynamic`.

vaibhavsagar:
To make this more pleasant to use, install [`cachix`](https://cachix.org/) and run `cachix use vaibhavsagar` before `nix-build release-8.6.nix` to use my binary cache.

PR #918, opened 2018-08-28
Drop support for GHC < 8.0

erikd:


erikd:
Don't merge this yet. Some more stuff to be cleaned out.

vaibhavsagar:
Thanks so much! Can you also remove the `binPkgDb` flag from `ihaskell.cabal` as well? That only exists to support GHC 7.10. Otherwise you can merge your changes as they are and I can do that.

erikd:
@vaibhavsagar Please feel free to merged this. Next task for me is `-Wall`.

vaibhavsagar:
Thanks so much!

PR #927, opened 2018-09-01
ihaskell-widgets: Make it compile with ghc 8.4

erikd:
GHC 8.4 allows value level operators (in this case list append, '++') to
be used at the type level, so that the singletons package for GHC 8.4
uses this operator whereas previous versions of singletons defined a ':++'
operator for type level append.


vaibhavsagar:
Thanks so much!

PR #929, opened 2018-09-03
Fix more warnings!

erikd:
The removal of the partial fields in `ihaskell` makes this an API breaking change. Should I do that change now or should wait until its time to release?


erikd:
And look! CI is happy!

vaibhavsagar:
This is great, thank you!

PR #931, opened 2018-09-05
ihaskell-widgets: Fix more warnings

erikd:
This file was giving a strange error/warning that had to be disabled.

After investigating this issue (https://ghc.haskell.org/trac/ghc/ticket/15601)
it turns out this was due to an wart in the way type level lists are handled
in GHC. Fortunately a workaround is available allowing the warning to be
re-enabled.


vaibhavsagar:
Thank you!

PR #937, opened 2018-09-15
Add ihaskell-graphviz to display graphs.

lucasdicioccio:
I've liberally copied a skeleton from ihaskell-gnuplot.

Example usage on my Mac:

screen shot 2018-09-15 at 22 23 47


vaibhavsagar:
This is great, thanks so much! I've wanted something like this for a long time :smile:.

lucasdicioccio:
My pleasure! I'll likely do a few more changes, but now that the footprint is here I guess the barrier to entry is low.

PR #939, opened 2018-09-16
Fix haddock generation.

### LINKED ISSUE #938: Error generating Haddocks
One thing I miss using IHaskell is having a Haddock on Hackage.

I've investigated a bit and it turns out there's a single error.

```
src/IHaskell/Types.hs:259:9: error: parse error on input ‘!’
        |
    259 |         ![DisplayData]
        |         ^
```

That's due to a mix of strictness annotation and inlined nested commented (`-- ^ Some comment`) at a field of a constructor of a sum-type. I bet it's worth trying to make a minimal repro case and file that against Haddock.

Patch trading off a bit of precision for an invaluable haddock dock in Hackage incoming in a few minutes.

lucasdicioccio:
Fixes #938 .

### DIFF
--- a/src/IHaskell/Types.hs
+++ b/src/IHaskell/Types.hs
@@ -249,14 +249,18 @@ instance ToJSON WidgetMethod where
   toJSON (CustomContent v) = object ["method" .= ("custom" :: Text), "content" .= v]
 
 -- | Output of evaluation.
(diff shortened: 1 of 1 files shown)


vaibhavsagar:
Thanks!

Issue #912, opened 2018-08-26
jupyterlab 0.34 notebook fails with irrefutable pattern


MMesch:
In `0.34`, the kernel stops with this Exception, when e.g. `print "3+4"` is executed in a new Haskell notebook:

```
Header: {"username":"","version":"5.2","session":"00cb7855-80d1-48e4-9b8d-407ffee191dd","msg_id":"2f2f7c14-daae-4211-8d72-870728919a73","msg_type":"execute_request","date":"2018-08-26T19:53:01.644284Z"}
Content: {"silent":false,"store_history":true,"user_expressions":{},"allow_stdin":true,"stop_on_error":true,"code":"print \"3+4\""}
ihaskell: src/IHaskell/IPython/Message/Parser.hs:71:5-81: Irrefutable pattern failed for pattern Just metadataMap

```

It doesn't fail when a code console is used instead of the notebook.

MMesch:
in `0.33` the same message looks like this:

```
Header: {"username":"","version":"5.2","session":"07e2d0c9-5c42-4ca3-9a67-4d6310624b06","msg_id":"89477dd2-606a-42ed-9864-c975f034e807","msg_type":"execute_request","date":"2018-08-26T20:00:21.663497Z"}
Content: {"silent":false,"store_history":true,"user_expressions":{},"allow_stdin":true,"stop_on_error":true,"code":"print \"3+4\""}
Message: PublishStatus {header = MessageHeader {identifiers = ["07e2d0c9-5c42-4ca3-9a67-4d6310624b06"], parentHeader = Just (MessageHeader {identifiers = ["07e2d0c9-5c42-4ca3-9a67-4d6310624b06"], parentHeader = Nothing, metadata = fromList [], messageId = UUID {uuidToString = "89477dd2-606a-42ed-9864-c975f034e807"}, sessionId = UUID {uuidToString = "07e2d0c9-5c42-4ca3-9a67-4d6310624b06"}, username = "", msgType = ExecuteRequestMessage}), metadata = fromList [], messageId = UUID {uuidToString = "c3be2f2a-938d-426b-b442-327de3bc6fe9"}, sessionId = UUID {uuidToString = "07e2d0c9-5c42-4ca3-9a67-4d6310624b06"}, username = "", msgType = StatusMessage}, executionState = Busy}
Sent: "{\"execution_state\":\"busy\"}"
[...]
```

MMesch:
identified the issue:

```
[...]

PR #940, opened 2018-09-17
update jupyterlab installation procedure in README

MMesch:
this small update makes the README correct until https://github.com/gibiansky/IHaskell/issues/912 is fixed


vaibhavsagar:
Thanks!

PR #825, opened 2018-02-16
[WIP] Use newer Hoogle

vaibhavsagar:
I was hoping this would be as easy as changing `queryUrl` but I guess not.

PR #953, opened 2018-10-24
Use hoogle v5.

dispanser:
Hoogle v5 uses a different json representation for the data, requires a
new URL and also contains HTML markup inside the json fields.

Based on #825, but handles the markup by removing it, and also adds a test or two for proper parsing of the hoogle response.

Note that we don't remove the markup from the doc string, but that actually looked OK in my tests.

### DIFF
--- a/ihaskell.cabal
+++ b/ihaskell.cabal
@@ -164,6 +164,7 @@ Test-Suite hspec
         IHaskell.Test.Completion
         IHaskell.Test.Util
         IHaskell.Test.Parser
+        IHaskell.Test.Hoogle
     default-language: Haskell2010
     build-depends:
         base,
@@ -178,6 +179,7 @@ Test-Suite hspec
         directory,
         text,
         shelly,
+        raw-strings-qq,
         setenv
 
 source-repository head
--- a/src/IHaskell/Eval/Hoogle.hs
+++ b/src/IHaskell/Eval/Hoogle.hs
@@ -1,24 +1,33 @@
-{-# LANGUAGE NoImplicitPrelude, FlexibleInstances, OverloadedStrings #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE OverloadedStrings #-}
 
 module IHaskell.Eval.Hoogle (
     search,
     document,
     render,
     OutputFormat(..),
-    HoogleResult,
+    HoogleResult(..),
+    HoogleResponse(..),
+    parseResponse,
     ) where
 
+import qualified Data.ByteString.Char8   as CBS
+import qualified Data.ByteString.Lazy    as LBS
+import           Data.Either             (either)
 import           IHaskellPrelude
-import qualified Data.ByteString.Lazy as LBS
-import qualified Data.ByteString.Char8 as CBS
 
+import           Data.Aeson
+import           Data.Char               (isAlphaNum, isAscii)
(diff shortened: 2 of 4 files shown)


vaibhavsagar:
Thanks so much! It looks like the response from Hoogle is also going to stdout, can that be removed/turned off? I'll also need you to update the sample notebook that we use for testing with the new Hoogle output so that the CI build passes.

dispanser:
I just disabled the debug trace. I'll look into the travis issue -- unfortunately I only now realized that this thing failed, did not get a notification.

I'm also not yet happy with the output rendering. While the patch strips off all the markup from the type signature (`item`), the associated docs still contain markup, and it doesn't really look pretty.

vaibhavsagar:
In that case, would you mind marking it `WIP`? I'd like to merge this once you're happy with it and the CI builds pass.

vaibhavsagar:
Thanks!

dispanser:
@vaibhavsagar I played around with the html representation a bit, here's three static html files that show how it looks

* hoogle v4: https://dispanser.de/ihaskell-hoogle-v4.html
* hoogle v5: https://dispanser.de/ihaskell-hoogle.html
* proposal: https://dispanser.de/ihaskell-hoogle-edited.html

In all fairness, the existing implementation doesn't look particularly pretty to begin with, so the bar is pretty low here :-).  The current solution suffers from the fact that ihaskell attempts to parse the doc string, which it assumes to be some ascii (interpreting `>` markers as code blocks).

I would like to aim for the third one, where the doc representation sent by hoogle (containing its html markup) is more or less literally put into the `hoogle-doc` `div`.

It would be even better to reuse the code highlighting and display mechanisms of the notebook itself, since it's just code interleaved with text, but I have now idea how to achieve that.

What do you think?

vaibhavsagar:
I like your proposed HTML representation the best. I don't have any idea how to reuse the code highlighting in the notebook either :frowning_face:, maybe we can address that in a follow-up PR when we know more?

dispanser:
@vaibhavsagar:

I have problems fixing the travis build. I attempted to re-run the `IHaskell.ipynb` notebook and safe it afterwards, is that the correct approach?

Here's the diff:

```
<        "null"
---
>        "Null"
1020c1020
<        "true"
---
>        "Bool True"
1029,1032c1029
<        "{\n"
<        "    \"x\": 3,\n"
<        "    \"y\": 2\n"
<        "}"
---
>        "Object (fromList [(\"x\",Number 3.0),(\"y\",Number 2.0)])"
```

The diffs just seem to be representational issues. I suspect this could be related to the fact that my local build is based on `release-8.4.nix`, however I can't get `release.nix` or `release-8.2.nix` to build (I'm on NixOS 18.09).

I could probably apply the diff above to fix the tests, but that definitely feels like cheating :-).

Also, two travis runs are cancelled due to too much log files generated, I can't see how this could possibly related to the PR.

vaibhavsagar:
That section of the notebook is meant to test `ihaskell-aeson` which isn't enabled by default with the Nix approach, so in this case I think the right thing to do is to apply the diff. Not sure about the log file issue, I haven't seen that recently.

dispanser:
Thanks, got it fixed by adding `ihaskell-aeson` to the packages. 

IMHO it's ready to be merged now. I see some potential improvements in the presentation, but that is independent from switching to hoogle v5.

If you know of more low-hanging fruit (beginner friendly issues), let me know :smile:

PR #972, opened 2018-11-16
Fix protocol communication with jupyter-lab.

qnikst:


vaibhavsagar:
If there is a test failure due to the new Hoogle integration, please rebase on the latest `master`, which I hope will fix this issue for the time being.

qnikst:
pushed, let's see. I've checked the build only on  `release.nix`.

PR #974, opened 2018-11-17
Remove duplicate module `IHaskellPrelude`.

### LINKED ISSUE #971: Duplicate module IHaskellPrelude
The module `IHaskellPrelude` exists twice, with identical content: `main/IHaskellPrelude.hs`, `src/IHaskellPrelude.hs`.

I find this confusing (and `ghcid` does not like it either with my setup). Is there any reason to not just export the module from the library and reuse it in the executable? This change seems to work for me locally, but maybe I'm overlooking something.

```
Error: Multiple files use the same module name:
       * IHaskellPrelude found at the following paths
         * /home/pi/wip/data-haskell/IHaskell/main/IHaskellPrelude.hs (ihaskell:exe:ihaskell)
         * /home/pi/wip/data-haskell/IHaskell/src/IHaskellPrelude.hs (ihaskell:lib)
```

dispanser:
remove module from `main`, and promote from `other-modules` to
`exported-modules` in `library`.

Tested with release.nix,  `stack build` and `ghcid`.

Not possible to test on MacOS / Ubuntu.

Fixes #971.


vaibhavsagar:
Thanks!

PR #1003, opened 2019-02-08
ihaskell-widgets: Port to vinyl-0.9

ijaketak:


vaibhavsagar:
Thanks for this PR! Happy to accept if the tests pass. I should however point out that `ihaskell-widgets` is incompatible with the latest `ipywidgets` because of https://github.com/JuliaGizmos/Interact.jl/issues/176.

PR #1007, opened 2019-02-19
stack.yaml: enable ihaskell-plot

vaibhavsagar:


vaibhavsagar:
`plot` doesn't build on GHC 8.6 yet.

PR #1008, opened 2019-02-21
Attempt to set error status

vaibhavsagar:


SKogelnik:
After testing this it seems like setting the output_type after an error occurs is working properly already. 

Is it only the checks failing, because the new notebook gets compared with the example notebook IHaskell.ipynb, where the output_type is still set to "display_data"?

vaibhavsagar:
I haven't figured out how to set the output_type for parse errors and the error output is black instead of red. It seems like the way IPython does this is by using janky unicode escaping but my first attempt at replicating that in IHaskell failed because GHC does not recognise that character sequence as valid. I will update the notebook though.

vaibhavsagar:
I also don't like my current solution of passing a boolean around, and I'd like to refactor it slightly so that the `ErrorOccurred` type is used instead.

SKogelnik:
For me the `output_type` after a Parse Error is already set to `error`:

```
"cells": [
  {
   "cell_type": "code",
   "execution_count": 2,
   "metadata": {},
   "outputs": [
    {
     "ename": "",
     "evalue": "",
     "output_type": "error",
     "traceback": [
      "Parse error (line 1, column 5): parse error (possibly incorrect indentation or mismatched brackets)"
     ]
    }
   ],
   "source": [
    "asd>"
   ]
  },
  {
   "cell_type": "code",
   "execution_count": null,
   "metadata": {},
   "outputs": [],
   "source": []
  }
 ]
```

Thank you for addressing this issue by the way.

vaibhavsagar:
Ah, I was thinking of nonexistent language extensions, e.g.

```haskell
:set -XBanana
```

Ideally this would be an error too, but it isn't yet :frowning_face:

vaibhavsagar:
I actually think this is in a mergeable state now.

Issue #187, opened 2014-02-04
SVG plots do not display axis values < 1 correctly


DougBurke:
It looks like there's a font/glyph issue if you create a plot in SVG format where the axis values are < 1; the decimal point is rendered as 5! If you switch to PNG format this goes away, so I'm wondering if there's something in the IHaskell environment/set-up that causes this problem. I tried some simple investigation to try and replicate this by using Chart directly from Haskell, but I'm not convinced I'm testing the same code as used by IHaskell here.

I don't know if this is just on my machine or something more general. I've uploaded a notebook showing the problem at https://gist.github.com/DougBurke/8804708 and you can view it at http://nbviewer.ipython.org/gist/DougBurke/8804708 .

PR #1009, opened 2019-03-01
Remove conversion of svg to html for generic usage.

GregorySchwartz:
On my setup it works with both browser and emacs front-ends.

### DIFF
--- a/src/IHaskell/Publish.hs
+++ b/src/IHaskell/Publish.hs
@@ -6,7 +6,6 @@ module IHaskell.Publish
 import           IHaskellPrelude
 
 import qualified Data.Text as T
-import qualified Data.Text.Encoding as E
 
 import           IHaskell.Display
 import           IHaskell.Types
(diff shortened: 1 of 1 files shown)


vaibhavsagar:
This is great, thanks!

GregorySchwartz:
It seems as though https://github.com/gibiansky/IHaskell/issues/187 was the reason for embedding. If this is the case, I don't think embedding is the answer, rather replacing the glyph like in https://github.com/jupyter/notebook/issues/333. Is it possible to get the cell id per call? If so, we can pass that id to the MimeSvg and just add those ids to all glyphs, clips, images, and gradients as well like they do in IJulia.

GregorySchwartz:
IJulia uses a timestamp as an id. The publish function already is in the IO monad, so that would be a simple solution. I can try to add it in. Should this merge be reverted in the meantime?

vaibhavsagar:
Sure, I've reverted it for now.

Issue #877, opened 2018-06-29
PNG output rather than SVG in emacs-jupyter


GregorySchwartz:
I'm not sure if this is an IHaskell issue or an ob-ipython issue, but with `:opt svg` the output is still a png file.

```
#+HEADER: :results raw drawer
#+BEGIN_SRC jupyter-haskell
:extension NoMonomorphismRestriction FlexibleContexts TypeFamilies
:opt svg

import Diagrams.Prelude
import qualified Diagrams.Prelude as D
import Diagrams.Backend.SVG

-- By Brent Yorgey
-- Draw a Sierpinski triangle!
sierpinski 1 = eqTriangle 1
sierpinski n =     s
                  ===
               (s ||| s) D.# centerX
  where s = sierpinski (n-1)

-- The `diagram` function is used to display them in the notebook.
diagram $ sierpinski 4
            D.# centerXY
            D.# fc black
          `atop` square 10
                   D.# fc white
#+END_SRC
```

vaibhavsagar:
Are you still experiencing this? It looks like the [relevant code](https://github.com/gibiansky/IHaskell/blob/ff046bf2c4c1c1c4efb54a692c4f20c9d5c113f0/ihaskell-display/ihaskell-diagrams/IHaskell/Display/Diagrams.hs#L18-L21) outputs both PNG and SVG.

GregorySchwartz:
I am. I do want to reiterate that it's through https://github.com/dzop/emacs-jupyter now. It works on the browser notebook, but not in the org-mode run. However, I don't see why it would be any different. Using the `jupyter-python` code blocks, I can get both `png` by default and `svg` with `%config InlineBackend.figure_format = 'svg'`, so I don't see how it would be different. Are there inconsistencies with reporting `svg` and `png` between these kernels?

GregorySchwartz:
Okay, see https://github.com/dzop/emacs-jupyter/issues/16#issuecomment-465728972 for the reason and a possible solution. It appears to be because the ihaskell kernel embeds the image in html. A possible solution is to provide it as a `image/svg+xml` similar to what IJulia and IPython do (which he says is according to the Jupyter spec).

vaibhavsagar:
I'm not sure what you mean by "embeds the image in html". Would you be able to point me to the relevant part of the code where this is happening?

GregorySchwartz:
I tried searching and I'm having difficulty finding where this information gets displayed in the code.

[...]

PR #1010, opened 2019-03-04
Make unique labels using a timestamp for svg elements.

GregorySchwartz:
Addresses https://github.com/gibiansky/IHaskell/issues/877 and issues raised in https://github.com/gibiansky/IHaskell/pull/1009.

### DIFF
--- a/ihaskell.cabal
+++ b/ihaskell.cabal
@@ -82,6 +82,7 @@ library
                        strict               >=0.3,
                        system-argv0         -any,
                        text                 >=0.11,
+                       time                 >= 1.8,
                        transformers         -any,
                        unix                 >= 2.6,
                        unordered-containers -any,
--- a/ihaskell.cabal
+++ b/ihaskell.cabal
@@ -82,7 +82,7 @@ library
                        strict               >=0.3,
                        system-argv0         -any,
                        text                 >=0.11,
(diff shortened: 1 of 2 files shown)


vaibhavsagar:
Can you rebase on top of current `master` please? Also, did you consider using something like [`Data.Unique`](http://hackage.haskell.org/package/base-4.12.0.0/docs/Data-Unique.html) instead of the timestamp?

GregorySchwartz:
While `Data.Unique` is a good idea, it would not be preserved between sessions, so it would become non-unique if saving and reopening. The best idea would be to use the cell number, but I don't know how to get that. So if it's cell `[34]` then the results of that cell would always be `34` for the label.

GregorySchwartz:
Rebased.

GregorySchwartz:
Looks like the check failed because of `stack-8.0`. I'm changing `getSystemTime` to `getCurrentTime` and using the date + picoseconds versus seconds from unix epoch as a workaround.

GregorySchwartz:
Used earlier version of time for `stack-8.0`. The `id` label, using date and picoseconds, contains "-" and "." characters, but it seems to meet specs so I left it as is.

vaibhavsagar:
Thanks!

PR #1011, opened 2019-03-05
Change extend to override in release.nix

imalsogreg:
`override` doesn't compose well with `extend` (https://github.com/NixOS/nixpkgs/issues/26561). This PR changes the use of `extend` to `override` (and simultaneously sneaks in a `dontCheck` wrapper for `inline-r`, is this ok?)


vaibhavsagar:
Thanks!

vaibhavsagar:
Is there a reason you didn't use the solution in the linked issue?

```nix
haskellPackages.override (old: {
  overrides = lib.composeExtensions (old.overrides or (_: _: {})) (self: super: {
    # ...
  });
})
```

imalsogreg:
@vaibhavsagar Good catch. No reason - it was an oversight! I see you have since fixed, thank you :)

PR #1021, opened 2019-04-20
labextension npm dependency bounds loosen

### LINKED ISSUE #1020: IHaskell for docker-stacks/jupyter
I'm trying to get IHaskell built on the [jupyter/docker-stacks/minimal-notebook](https://hub.docker.com/r/jupyter/minimal-notebook) and I would appreciate any advice.

With this `Dockerfile`:

```
FROM jupyter/minimal-notebook:7d427e7a4dde

USER root

RUN \
    apt-get update && \
    apt-get install -y --no-install-recommends \
        python3-pip \
        git \
        libtinfo-dev \
        libzmq3-dev \
        libcairo2-dev \
        libpango1.0-dev \
        libmagic-dev \
        libblas-dev \
        liblapack-dev \
        g++ \
        gcc \
        libc6-dev \
        libffi-dev \
        libgmp-dev \
        make \
        xz-utils \
        zlib1g-dev \
        git \
        gnupg \
        netbase \
        curl \
        && \
    rm -rf /var/lib/apt/lists/*

RUN \
    curl -sSL https://get.haskellstack.org/ | sh

USER $NB_UID

ENV PATH $(stack path --local-install-root)/bin:$(stack path --snapshot-install-root)/bin:$(stack path --compiler-bin):/home/${NB_USER}/.local/bin:${PATH}

COPY --chown=jovyan:users IHaskell/ IHaskell/
RUN fix-permissions /home/jovyan

RUN \
    cd IHaskell && \
    stack install --fast && \
    ihaskell install --stack

RUN npm install -g typescript

RUN \
    cd IHaskell/ihaskell_labextension && \
    npm install && \
    npm run build && \
    jupyter labextension install .
```

Try to build the Jupyter Notebook:

```
git clone git@github.com:gibiansky/IHaskell.git
docker build --rm=false -t minimal-notebook-ihaskell:latest .
```

And it errors out this way:

```
Errored, use --debug for full output:
ValueError: 
"ihaskell_jupyterlab@0.0.6" is not compatible with the current JupyterLab
Conflicting Dependencies:
JupyterLab              Extension        Package
>=0.19.1 <0.20.0        >=0.17.1 <0.18.0 @jupyterlab/application
>=0.19.1 <0.20.0        >=0.17.2 <0.18.0 @jupyterlab/apputils
>=0.19.2 <0.20.0        >=0.17.1 <0.18.0 @jupyterlab/notebook
```

jamesdbrock:
Relax upper version bounds on the jupyterlab alpha dependencies
for ihaskell_labextension. It looks to me like these version bounds
were intended to allow the entire major alpha 0.\*.\* version range, but
the the prefix caret only includes minor versions if the major version
is 0.

Let's set the upper bound to 2.0.0, because according to SemVer, that
is the next version number which we would expect to have breaking
changes.

Resolves #1020

### DIFF
--- a/ihaskell_labextension/package.json
+++ b/ihaskell_labextension/package.json
@@ -30,10 +30,10 @@
     "watch": "tsc -w"
   },
   "dependencies": {
-    "@jupyterlab/application": "^0.17.1",
-    "@jupyterlab/apputils": "^0.17.2",
-    "@jupyterlab/docregistry": "^0.17.1",
-    "@jupyterlab/notebook": "^0.17.1",
+    "@jupyterlab/application": "0.17.1 - 2.0.0",
+    "@jupyterlab/apputils": "0.17.2 - 2.0.0",
+    "@jupyterlab/docregistry": "0.17.1 - 2.0.0",
+    "@jupyterlab/notebook": "0.17.1 - 2.0.0",
     "@jupyterlab/services": "^3.0.1",
     "@phosphor/disposable": "^1.1.2"
   },


vaibhavsagar:
Wonderful, thanks so much! The test failures are because the integration tests currently depend on the ordering of Hoogle output, which changes frequently.

PR #1033, opened 2019-05-21
Fix compliance with Jupyter 5.0 spec for idle messages

thomasjm:
IHaskell says it's on Jupyter protocol version 5.0, but it isn't properly following this note:

```
Changed in version 5.0: Busy and idle messages should be sent before/after handling every request, not just execution.
```

(from https://jupyter-client.readthedocs.io/en/stable/messaging.html#general-message-format)

This PR fixes IHaskell to send busy and idle messages after every request.


vaibhavsagar:
Thanks so much!

PR #1038, opened 2019-06-03
Dockerfile refactor

jamesdbrock:
This Dockerfile gives us more control and isolation of dependencies
during development.

Base image `FROM ubuntu` instead of `fcpo/stack-build`.  We will use the
Stack and GHC on the host system, don't need it installed in the image.
The image is just for isolating the system package dependencies. Stack
will isolate the Haskell package dependencies (and the pip and npm
dependencies).

This gives is a single authority for the resolver, which is
the `stack.yaml`. We no longer need to manually make the `stack.yaml`
and `Dockerfile` resolvers agree.

Don't install `jupyter` in the Docker image, we'll use `stack --docker` to
install it later. This allows developers to choose the version of the jupyter
dependencies.

Add the `docker` section to `stack.yaml`, disabled by default.

__Motivating example:__ ipywidgets version 7 has changed its protocol.
We want to upgrade ihaskell-widgets, which works only on ipywidgets
version 6. To install ipywidgets version 6, we need to downgrade Python to 3.5.
We can use `stack --docker` to isolate a development environment in
which ihaskell-widgets is working, and then upgrade it to the latest
versions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,13 @@ extension to get syntax highlighting with:
 jupyter labextension install ihaskell_jupyterlab
 ```
 
+Run Jupyter.
+
+```bash
+stack exec jupyter -- notebook
+```
+
+
 ## Mac
 
 You need to have [Homebrew](https://brew.sh) installed. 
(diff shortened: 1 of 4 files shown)


vaibhavsagar:
This looks great, thanks!

jamesdbrock:
@vaibhavsagar You're admirably fast with these merges.

vaibhavsagar:
Thanks! I figure I can always revert a merge if it turns out it was a bad idea.

PR #1039, opened 2019-06-03
rm ihaskell-display/ihaskell-*/stack.yaml

jamesdbrock:
Andrew Gibiansky created all these package `stack.yaml`s with this commit:

> commit f9846b6bd9f162b763cc53662e0db357b2aec03a
> Author: Andrew Gibiansky <andrew.gibiansky@gmail.com>
> Date:   Tue Jun 7 17:26:47 2016 -0700
>
>     Separate stack.yaml for each package for simplicity

The resolvers for all these stack yamls originally agreed with the
root stack.yaml, but now they're way off, apparently out of neglect.
Let's just delete them.

### DIFF
--- a/ihaskell-display/ihaskell-aeson/stack.yaml
+++ b/ihaskell-display/ihaskell-aeson/stack.yaml
@@ -1,4 +1,0 @@
-flags: {}
-packages:
-- '.'
-resolver: lts-9.21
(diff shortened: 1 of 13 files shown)


vaibhavsagar:
Great idea, thanks!

PR #1042, opened 2019-06-05
Stable nbconvert acceptance test

### LINKED ISSUE #1041: acceptance testing, no Hoogle
We should factor the acceptance test out of the `.travis.yaml` into a separate script that is called from `.travis.yaml`.

```
stack exec -- jupyter nbconvert --to=notebook --execute --allow-errors --stdout notebooks/IHaskell.ipynb > ~/ihaskell-out.ipynb
 # Images are rendered differently on different systems, so filter them out in the comparison
diff <(egrep -v 'image/png|version|pygments' ~/ihaskell-out.ipynb) <(egrep -v 'image/png|version|pygments' notebooks/IHaskell.ipynb)
```

Also we should change this acceptance test so that it doesn't depend on Hoogle output, or anything on the internet. See #1034 

We could also add some `console` acceptance tests, so for example check the output of something like:

```
echo 'let 1+1=3 in 1+1' | stack exec jupyter -- console --kernel=haskell --log-level=0 --simple-prompt
```

And while we're at it, we should probably make a `test` subdirectory for this stuff.

jamesdbrock:
Change the `nbconvert` test so that it converts a notebook with no
Hoogle (which changes over time) and no images (which differ on
different systems).

This should reduce the number of surprise Travis test failures.

Resolves #1041

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -78,10 +78,7 @@ matrix:
               # Run the notebook to regenerate the outputs, then compare the new notebook to the old one.
               stack install --fast
               stack exec -- ihaskell install --stack
-              stack exec -- jupyter nbconvert --to=notebook --execute --allow-errors --stdout notebooks/IHaskell.ipynb > ~/ihaskell-out.ipynb
-
-              # Images are rendered differently on different systems, so filter them out in the comparison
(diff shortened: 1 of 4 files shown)


vaibhavsagar:
Sure, let's do it this way.

Issue #1055, opened 2019-06-25
Evaluation tests fail on nix master


dylex:
I'm trying to build ihaskell from the current release.nix against the current nixpkgs master (in an overlay):

```
ihaskell = import IHaskell/release.nix {
  nixpkgs = self;
  compiler = "ghc865";
}
```

but all of the evaluation tests seem to fail with

```
  11) Code Evaluation evaluates directives
       uncaught exception: SourceError
       Ambiguous module name ‘Language.Haskell.TH’: it was found in multiple packages: ghc-lib-parser-8.8.0.20190424 template-haskell-2.14.0.0
```

I'm trying with ghc864 now...

dylex:
ghc864 has the same issue. We do have a few other changes in our overlay, but nothing that seems like it would cause this.

vaibhavsagar:
Can you share the expression you're using to build IHaskell? I suspect this might be because `hlint` depends on `ghc-lib-parser` now.

dylex:
See https://github.com/flatironinstitute/nix-modules/blob/ihaskell/overlay.nix#L319
And just `nix-build -A ihaskell .` in that repo (which should essentially be nixpkgs 9638d85ff6b with minor tweaks). Something definitely pulled in ghc-lib-parser.

vaibhavsagar:
Can you try building with `hlint-2.1.17`?

dylex:
Switching to hlint 2.1.17 and removing the ghc-lib-parser dependency does indeed fix it.  (Switching to 2.1.17 but leaving the unused nix dependency, it's still broken.)

vaibhavsagar:
This has been fixed by making IHaskell hide `ghc-lib` and `ghc-lib-parser` from the package database. Please try again with the latest commit of IHaskell, and reopen if this still doesn't fix your issue.

PR #1057, opened 2019-06-27
src/IHaskell/Eval/Evaluate.hs: hide ghc-lib{,-parser}

vaibhavsagar:


vaibhavsagar:
This addresses #1055.

PR #1063, opened 2019-07-01
Help more Windows 10 users use IHaskell, via WSL

mpilgrem:
Update `README.md` to reflect the advice of Arvind Devarajan at https://blog.ramdoot.in/ihaskell-on-windows-c549e6442262

For Windows 10 users possibly unfamiliar with Linux, things are taken step-by-step.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,11 +16,6 @@ The [wiki](https://github.com/gibiansky/IHaskell/wiki) also has more extensive d
 
 ### Interactive In-Browser Notebook
 
-
-**Note:** IHaskell does not support Windows. To use on Windows, install
-Virtualbox, install Ubuntu or another Linux distribution, and proceed with the
-install instructions.
-
 # Installation
 
 ## Linux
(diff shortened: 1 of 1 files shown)


mpilgrem:
I think 'unnecessarily' depends on your perspective. If you are a developer familiar with Linux, including via WSL, then I agree that 'use WSL, see Linux instructions' may be all that you need.

If you are a Windows 10 user who has had little or no contact with Unix-like operating systems, and even may not know that WSL exists, then step-by-step, with a little additional explanation for the 'Ubuntu app' steps, under the heading 'Windows' may be appreciated. I appreciated the blog entry that I referenced - it 'unlocked' IHaskell for me - and thought it would be useful for the nub of the advice to be available in the repository itself.

vaibhavsagar:
I don't see it as the responsibility of the IHaskell `README` to educate Windows users on exactly how to use WSL, and I would in fact prefer "use WSL, see Linux instructions", which I think is the nub of the advice. You're right that this is a matter of perspective, but I'm not going to merge a copy of essentially the same instructions we already provide.

mpilgrem:
I will re-propose, then, an abbreviated version that cross-refers to the Linux instructions. It may be a day or two before I can turn to that.

vaibhavsagar:
Thanks! I'm in no rush.

PR #1065, opened 2019-07-06
src/IHaskell/Eval/Evaluate.hs: support GHC 8.8

vaibhavsagar:


vaibhavsagar:
This is available on https://vaibhavsagar.cachix.org to play with :smiley:.

PR #1068, opened 2019-07-06
src/IHaskell/Eval/Evaluate.hs: unhide ghc-lib{,-parser} after hiding

vaibhavsagar:


vaibhavsagar:
This doesn't work :sob:.

PR #1070, opened 2019-07-11
ihaskell_labextension for JupyterLab 1.0.0

jamesdbrock:
If we remove the upper version bound for the `jupyterlab/services`
dependency, then `ihaskell_labextension` works fine with __JupyterLab 1.0.0__.

https://github.com/jupyterlab/jupyterlab/commit/e2fd4c8841a7393f9131fb6b6e8252864a8bd351


vaibhavsagar:
Great, thanks!

PR #1075, opened 2019-07-23
Bind ZMQ Sockets to IP configured in Kernel Profile

zampino:
Hello @gibiansky 👋,

the proposed changes will let ZMQ sockets respect the address configured in the kernel profile. This is useful in the case when external connectivity is required (i.e. addresses other than localhost).

[Kernel connection file specification](https://jupyter-client.readthedocs.io/en/stable/kernels.html#connection-files) from Jupyter suggests it should be configurable and the Profile type seems to confirm this:

https://github.com/gibiansky/IHaskell/blob/50116ade4ebeb93198a55b6b7f8e4a2e4ed008f8/ipython-kernel/src/IHaskell/IPython/Types.hs#L71-L81

Most kernels we've worked so far including IPython, IJulia, IRkernel also support this approach.


vaibhavsagar:
Thank you!

PR #1090, opened 2019-08-16
ipython-kernel: support custom mimetypes

vaibhavsagar:


rehno-lindeque:
Might be worth thinking about the case where `dataStr` could be a json value, similar to say https://github.com/jupyter-widgets/ipywidgets/blob/3023b3ed24fbb172eb44643798a603aaed15ae76/packages/schema/messages.md#displaying-widgets-1

E.g. doesn't seem very elegant, but perhaps:

```
displayDataToJson (DisplayData (MimeCustom custom) dataStr) =
    custom .= fromMaybe (String "") (decodeStrict (Text.encodeUtf8 dataStr) :: Maybe Value)
```

vaibhavsagar:
I might be misunderstanding the linked documentation, but it seems like we need to support `application/vnd.jupyter.widget-view+json` as a separate mimetype. I can't see how the custom mimetype logic would be able to handle arbitrary JSON mimetypes correctly.

rehno-lindeque:
That's right yes, I do agree `application/vnd.jupyter.widget-view+json` should be a directly supported mimetype.

However, it seems plausible to me that there are more unknown situations where json contents is required to go along with the custom mimetype. Can't say I'm familiar enough with jupyter to say if this is common or not with custom mime types... (?)

(Side note: It doesn't apply to any other format - say XML for example - because XML would always be packed into the message as a text string. But with JSON there's this odd case of it being sent inline in the message rather than being packed into a json string.)

DougBurke:
I think you'd also want something like the following in `src/IHaskell/Display.hs`

```
custom ::
  T.Text
  -- ^ The mime type
  -> String 
  -- ^ The payload
  -> DisplayData
custom mtype = DisplayData (MimeCustom mtype) . T.pack
```

vaibhavsagar:
Is it important that the payload is a `String`, or can it be `Text` also?

DougBurke:
I was just mirroring what the other constructor-functions take, as they all seem to be `String` (e.g. `vega`, `markdown`, ...). I have no objection to it being a `Text`

DougBurke:
What needs to be done to move this forward?

vaibhavsagar:
Whoops, this slipped off my radar. I've made the change you requested to add a `custom` constructor, let me know if that looks good and if so, I can merge.

DougBurke:
It looks good to me (although I admit to not having actually tried it out yet). 

I think it would be helpful to add comments describing which argument to `custom` is the mimetype and which the payload (as an occasional user of the API I find the documentation for IHaskell rather sparse, so end up having to look at the source code).

DougBurke:
Thanks. Looking forward to trying this out .

PR #1092, opened 2019-09-01
README extra-deps vs packages

andrevdm:
As far as I can see using extra-deps to add dependencies is the correct with stack now. I belive its been that way for a while now so no need to keep the old docs around. Hopefully this helps others to get going a little faster.

Please let me know if this makes sense or if there is anything I can improve


andrevdm:
Nope looks like I'm still not quite following how to correctly add a dependency :(

Issue #510, opened 2015-06-09
Strange hlint message


aueelis:
I discovered some strange message from hlint in the Notebook while playing with some code.

When I type

```
sn = (round $ sqrt $ 2545)
```

hlint discovers the unnecessary brackets and dollar sign. Though, the error message is a bit wrong:

```
Redundant bracket
Found:
(round sqrt sqrt  2545)
Why Not:
round sqrt sqrt  2545

Redundant $
Found:
sqrt $ 2545
Why Not:
sqrt 2545
```

Obviously, the part `$ sqrt $` gets replaced by `sqrt sqrt`. Passing this line directly to hlint produces the desired output.
##

gibiansky:
Confirmed. Strange. I wonder if this has something to do with MathJax or something...

vaibhavsagar:
Closed by #1111.

PR #1111, opened 2019-10-18
IHaskell.Eval.Lint: escape $

vaibhavsagar:


vaibhavsagar:
Closes #510. I did a quick test with e.g. `&` and it didn't look like that needed to be escaped, but happy to revisit that if it becomes an issue.

PR #1126, opened 2019-11-25
Fix nix install instructions

srid:


vaibhavsagar:
Wow, those instructions are badly out-of-date. Thanks for bringing this up! Fixed in a442c0b6d4cf1fb17eaa3bf2827f04b21ad334bf.

PR #1142, opened 2020-02-11
fix typo in channel name

ttylec:
There was a typo in `ipython-kernel` that caused infinite loop when starting custom kernel using `easyKernel`: reply was send to wrong channel.


vaibhavsagar:
Thank you!

PR #1145, opened 2020-02-28
Add `buffers` field of the Wire Protocol

### LINKED ISSUE #1144: MessageHeader buffers
There is a way for the Jupyter frontend to communicate arbitrary binary blobs with the kernel, by the [`IMessage.buffers`](https://github.com/jupyterlab/jupyterlab/blob/b66bb79116285ccbd0876a6263ac2c0d6de27c33/packages/services/src/kernel/messages.ts#L277)

It would be nice if the IHaskell IPython kernel could send and receive `buffers`. I suppose they would be added here, in the [`IHaskell.IPython.Types.MessageHeader`](https://github.com/gibiansky/IHaskell/blob/9dd237ea0480302fd7a34c0eeef0fe97bca3ae1a/ipython-kernel/src/IHaskell/IPython/Types.hs#L149). 

 This new field would have the type `buffers :: [ByteString]`, just like `mhIdentifiers :: [ByteString]` which, by the way, stores the ASCII-encoded `mhSessionId` and `mhMessageId`.

Perhaps `buffers` was added in some latter Jupyter protocol version which IHaskell hasn't caught up with? I haven't investigated yet.

jamesdbrock:
Resolves #1144

### DIFF
--- a/ipython-kernel/ipython-kernel.cabal
+++ b/ipython-kernel/ipython-kernel.cabal
@@ -52,7 +52,8 @@ library
                        transformers    ,
                        unordered-containers,
                        uuid            ,
-                       zeromq4-haskell
+                       zeromq4-haskell ,
+                       parsec
 
 -- Example program
 executable simple-calc-example
--- a/ipython-kernel/src/IHaskell/IPython/EasyKernel.hs
+++ b/ipython-kernel/src/IHaskell/IPython/EasyKernel.hs
@@ -125,7 +125,7 @@ createReplyHeader parent = do
       err = error $ "No reply for message " ++ show (mhMsgType parent)
 
   return $ MessageHeader (mhIdentifiers parent) (Just parent) (Metadata (HashMap.fromList []))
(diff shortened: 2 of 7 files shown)


vaibhavsagar:
Thanks!

vaibhavsagar:
Out of curiousity: why `parsec` instead of `attoparsec`?

jamesdbrock:
@vaibhavsagar Because the type of the "input stream" that we're parsing here is `[ByteString]`, where the individual tokens are of type `ByteString`. Parsec and Megaparsec can parse weird input streams like that, but Attoparsec can only parse two types of input streams: `Text` and `ByteString`. And Parsec is already a [dependency of IHaskell](https://github.com/gibiansky/IHaskell/blob/f540feb155353427427e883462de44b0c100d64e/ihaskell.cabal#L76), but Megaparsec isn't, so I figured adding Parsec as a dependency of ipython-kernel was the least-dependency-adding thing to do.

Thanks as always for quick merging!

vaibhavsagar:
That makes sense, thanks for the clarification and the PR!

PR #1150, opened 2020-03-25
Support GHC 8.10

vaibhavsagar:


vaibhavsagar:
I worked around https://github.com/NixOS/nixpkgs/issues/79441 by commenting out the line that calls `ghc-pkg recache` and the resulting notebook ran without any noticeable issues, so I think this is ready to merge once the tests pass.

PR #1151, opened 2020-03-30
Upgrade ihaskell_labextension to JupyterLab 2

jamesdbrock:
https://jupyterlab.readthedocs.io/en/stable/developer/extension_migration.html

__ihaskell_labextension__ is not buildable against the latest JupyterLab.

This upgrade of __ihaskell_labextension__ to JupyterLab 2.0.2 seems to work, on my machine, and passes the acceptance test. What problems could this cause that I haven't thought about?

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -10,7 +10,7 @@
 #     docker build --build-arg BASE_CONTAINER=ubuntu:16.04 -t ihaskell-dev .
 #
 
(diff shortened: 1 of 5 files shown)


vaibhavsagar:
Seems fine to me, I'm happy to merge this and work out any issues that arise when we become aware of them.

GTrunSec:
With Command `jupyter labextension install ihaskell_labextension`
https://registry.npmjs.org/ihaskell_labextension Not found the package.

GTrunSec:
it looks like doesn't work on 2.0.2 version

image

PR #1154, opened 2020-04-16
IHaskell.ipynb replace dead img src

jamesdbrock:
I don't know what image was there before, but it's gone now. Let's put in the
Haskell logo from haskell.org.


vaibhavsagar:
Thanks!

Issue #1173, opened 2020-05-16
custom mimetype support - add direct JSON


DougBurke:
Can we add some way to pass JSON through displayDataToJson for the new custom mimetype? Rarndom thoughts are

a) add a separate mimetype - e.g.  `MimeCustomJSON Text` - which can then be used in `disolayDataToJson` in a similar way to how `MimeJson/MmeVegaite/MimeVega` is handled 

b) or change the MimeCustom type to treat the value in `displayDataToJson` directly as an Aeson Value

The bit I'd like to be able to handle is to pass a custom mimetype with a Value, not a String, in the followig.

```
-- from https://github.com/gibiansky/IHaskell/blob/master/ipython-kernel/src/IHaskell/IPython/Types.hs

displayDataToJson :: DisplayData -> (Text, Value)
displayDataToJson (DisplayData MimeJson dataStr) =
    pack (show MimeJson) .= fromMaybe (String "") (decodeStrict (Text.encodeUtf8 dataStr) :: Maybe Value)
displayDataToJson (DisplayData MimeVegalite dataStr) =
    pack (show MimeVegalite) .= fromMaybe (String "") (decodeStrict (Text.encodeUtf8 dataStr)
:: Maybe Value)
displayDataToJson (DisplayData MimeVega dataStr) =
    pack (show MimeVega) .= fromMaybe (String "") (decodeStrict (Text.encodeUtf8 dataStr) :: Maybe Value)
displayDataToJson (DisplayData mimeType dataStr) =
  pack (show mimeType) .= String dataStr

```

vaibhavsagar:
What is the problem you're trying to solve here? Could you encode the `Value` as a `Text` and use the existing data types? If it turns out to be necessary to add a new mimetype we can do that but I'm wary because each time it breaks backwards compatibility and complicates the interface.

DougBurke:
I am trying to test out the custom mimetype by changing

instance IHaskellDisplay VegaLiteLab where
  display (VLL vl) = let js = LT.unpack (encodeToLazyText (fromVL vl))
                     in pure (Display [vegalite js])

to

instance IHaskellDisplay VegaLiteLab where
  display (VLL vl) = let js = LT.unpack (encodeToLazyText (fromVL vl))
                         vegalite' = "application/vnd.vegalite.v4+json"
                     in pure (Display [custom vegalite' js])

and it does not work because the custom message gets encoded as a String,
whereas the vegalite message gets special cased to handle JSON. It is
[...]

PR #1176, opened 2020-05-29
Add a MimeCustomJson type

DougBurke:
The MimeCustom type encodes the payload as a JSON string, whereas
MimeCustomJson encodes it as a JSON value. An alternative would
have been to just convert the MimeCustom type to encode as a Value
directy, but I wanted to see how this works.

This is for #1173 and suffers from the issue discussed at https://github.com/gibiansky/IHaskell/issues/1173#issuecomment-629931984

On the plus side, using it I was able to add a custom Mime handler for an updated version of the VegaLite mime type (but only v3 and not v4 as I think that either requires updating to jupyter lab v2 or adding some mythical support to jupyter lab 1.2 to add the v4 mimetype)

### DIFF
--- a/ipython-kernel/src/IHaskell/IPython/Types.hs
+++ b/ipython-kernel/src/IHaskell/IPython/Types.hs
@@ -781,6 +781,7 @@ data MimeType = PlainText
               | MimeVegalite
               | MimeVdom
               | MimeCustom Text
+              | MimeCustomJson Text
   deriving (Eq, Typeable, Generic)
 
 -- Extract the plain text from a list of displays.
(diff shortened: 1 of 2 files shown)


DougBurke:
I've been away from IHaskell land for a while. Have there been any changes to this area of code or would this approach still work? I guess https://github.com/gibiansky/IHaskell/issues/1173#issuecomment-629931984 is still the problem?

vaibhavsagar:
Yes, more or less. I'm still not convinced it's a good idea to add an additional type that only permits a one-way conversion, but I'm willing to consider it if it makes your life significantly easier.

DougBurke:
As noted in https://github.com/gibiansky/IHaskell/issues/1173#issuecomment-866034312 I'm currently not working on this, but if anyone else is interested then please push ahead.

PR #1178, opened 2020-06-19
sendMessage mhMetadata

### LINKED ISSUE #1177: sendMessage metadata
https://github.com/gibiansky/IHaskell/blob/a992ad83702e55b774de234d77ffd2682d842682/ipython-kernel/src/IHaskell/IPython/ZeroMQ.hs#L347

`sendMessage`, instead of sending the `mhMetadata`, sends `"{}"`.

`receiveMessage` handles `mhMetadata` just fine. Here is where parsing metadata occurs, we just need the inverse of this.
https://github.com/gibiansky/IHaskell/blob/a992ad83702e55b774de234d77ffd2682d842682/ipython-kernel/src/IHaskell/IPython/Message/Parser.hs#L64

I will submit a PR.

jamesdbrock:
Send the header metadata with a message instead of ignoring the header
metadata and sending "{}".

The header metadata is handled correctly on receipt of a message. Why
was it ignored on sending? Probably no-one ever needed it.

Resolves #1177

### DIFF
--- a/ipython-kernel/src/IHaskell/IPython/ZeroMQ.hs
+++ b/ipython-kernel/src/IHaskell/IPython/ZeroMQ.hs
@@ -344,6 +344,6 @@ sendMessage debug hmackey sock msg = do
     hdr = header msg
     parentHeaderStr = maybe "{}" encodeStrict $ mhParentHeader hdr
     idents = mhIdentifiers hdr
-    metadata = "{}"
+    metadata = let Metadata mdobject = mhMetadata hdr in encodeStrict mdobject
     content = encodeStrict msg
     headStr = encodeStrict hdr


jamesdbrock:
I'm puzzled by that one failing test. I'd like to re-run it.

vaibhavsagar:
The tests are flaky on GHC 8.0, I wouldn't worry about it.

vaibhavsagar:
Thanks!

vaibhavsagar:
I think it would be good to add a test that checks that this gets passed through properly.

jamesdbrock:
> I think it would be good to add a test that checks that this gets passed through properly.

Yes... that would definitely be good. I'm not sure how to write that test with normal Notebooks. I guess Notebooks generally don't use this part of the protocol, which is why it went unimplemented in __ipython-kernel__ for so long.

I've tested that this feature works for my proprietary JupyterLab extension. If I think of a way to test it with only stock Notebooks then I'll make a PR for that.

Thanks as always for fast merging!

jamesdbrock:
This is such an obscure and unimportant corner of the protocol that Andrew just wrote `"{}"` because he couldn't be bothered to write one line of code, lol.

PR #1182, opened 2020-08-12
Add mimetype to Kernel Info Reply message

### LINKED ISSUE #1181: Kernel Info Reply mime_type
The Kernel Info Reply message should have a `mime_type`, as specified in https://jupyter-client.readthedocs.io/en/stable/messaging.html#kernel-info

https://github.com/gibiansky/IHaskell/blob/cc427517a59bfc8e1014c7e02fe5cfe0ce7e5eef/ipython-kernel/src/IHaskell/IPython/Types.hs#L294-L301

https://github.com/krassowski/jupyterlab-lsp/issues/313#issuecomment-672058096

I think the `mime_type` should be `"text/x-ihaskell"`?

jamesdbrock:
The `"mimetype"` is a required field, and is needed by __jupyterlab-lsp__ and probably other things.

See https://github.com/krassowski/jupyterlab-lsp/issues/313

Resolves #1181

### DIFF
--- a/ipython-kernel/src/IHaskell/IPython/Types.hs
+++ b/ipython-kernel/src/IHaskell/IPython/Types.hs
@@ -281,6 +281,10 @@ instance FromJSON MessageType where
       _                     -> fail ("Unknown message type: " ++ show s)
   parseJSON _ = fail "Must be a string."
 
+-- | Kernel language info, see
+--
+-- * https://jupyter-client.readthedocs.io/en/stable/messaging.html#kernel-info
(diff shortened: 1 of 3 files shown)


vaibhavsagar:
Can you update the test notebook as well so that the test suite passes? Happy to merge after that.

jamesdbrock:
> update the test notebook as well so that the test suite passes?

Oh yeah, done and force-pushed.

jamesdbrock:
Resolves #1181

vaibhavsagar:
Thanks so much!

PR #1184, opened 2020-08-25
Add dynamic flag to load extra shared library

junjihashimoto:
This pr adds -dynamic flag for ihaskell executable file.
External libraries such as libstdc++ cannot be loaded dynamically because there is no dynamic flag.

I verified it here.
https://github.com/hasktorch/hasktorch-jupyter/blob/master/Dockerfile#L55-L56


vaibhavsagar:
Thanks!

PR #1185, opened 2020-08-25
Bump mimetypes of vega

junjihashimoto:
The latest jupyterlab does not support both vega-v2 and vegalite-v2.
https://github.com/jupyterlab/jupyterlab/blob/f0153e0258b32674c9aec106383ddf7b618cebab/packages/vega5-extension/src/index.ts#L33-L49

So it can not display the images using vega.

This pr dumps vega's version of mimetype to show vega-images.

### DIFF
--- a/ipython-kernel/src/IHaskell/IPython/Types.hs
+++ b/ipython-kernel/src/IHaskell/IPython/Types.hs
@@ -814,8 +814,8 @@ instance Show MimeType where
   show MimeMarkdown = "text/markdown"
   show MimeJavascript = "application/javascript"
   show MimeJson = "application/json"
-  show MimeVega = "application/vnd.vega.v2+json"
-  show MimeVegalite = "application/vnd.vegalite.v2+json"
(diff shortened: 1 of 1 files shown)


vaibhavsagar:
Thanks!

tscholak:
@vaibhavsagar many thanks for merging these fixes, that will unblock our efforts downstream with hasktorch!
can I ask you to cut a release and push this to hackage, so that it can be picked up by cabal and haskell.nix? :)

vaibhavsagar:
@tscholak I uploaded [`ipython-kernel-0.10.2.1`](https://hackage.haskell.org/package/ipython-kernel-0.10.2.1) and [`ihaskell-0.10.1.2`](https://hackage.haskell.org/package/ihaskell-0.10.1.2). I chose not to bump the bounds on `ihaskell` to require the new `ipython-kernel` since it should still work with older versions (except for the VEGA version change).

tscholak:
thank you!!

junjihashimoto:
Thank you for cooperating so kindly.

PR #1189, opened 2020-09-20
update ihaskell jupyterlab extension and align directory name

MMesch:
PR that brings the ihaskell jupyterlab extension up to date with https://www.npmjs.com/package/jupyterlab-ihaskell .

* it renames the directory
* it aligns the extension name with the one used on npmjs


vaibhavsagar:
Thank you!

PR #1213, opened 2021-02-28
Support for GHC 8.10

serras:
The main reason why this is important is that GHC 8.10.3 ships with a [fix](https://gitlab.haskell.org/ghc/ghc/-/issues/18446) that allows compilation in macOS Big Sur, which otherwise fails.


vaibhavsagar:
Thanks! Is there a reason you used `allow-newer` in `stack.yaml` instead of updating the bounds in `ihaskell.cabal`?

serras:
I found that some packages (not `ihaskell` directly) had strict bounds on `base`, which made it impossible to compile with GHC 8.10.

PR #1222, opened 2021-06-08
Add a stack-based hie.yaml for haskell-language-server

jamesdbrock:
We want to be able to clone IHaskell, fire up a
haskell-language-server-capable editor, and have it just work with no
extra configuration steps.

If we have Nix and Stack installed on our system, then with this commit
we get what we want. To make this work, we had to add a
special `stack-nix.yaml` which enables Stack Nix integration by default.

This is a bit awkward because now a `stack-nix.yaml.lock` file gets
generated and what should we do with that? Check it in? In this commit
we've decided to ignore it.

See also https://github.com/mpickering/hie-bios/issues/297


daviddavo:
I'm getting the following error with vscode
```
Couldn't figure out what GHC version the project is using: 
/home/davo/.config/Code - OSS/User/globalStorage/haskell.haskell/haskell-language-server-wrapper-1.1.0-linux --project-ghc-version exited with exit code 1: 
Found "/home/davo/Documents/IHaskell/hie.yaml" for "/home/davo/Documents/IHaskell/a" 
Module "/home/davo/Documents/IHaskell/a" is loaded by Cradle: 
Cradle {cradleRootDir = "/home/davo/Documents/IHaskell", cradleOptsProg = CradleAction: Stack} 
Failed to get project GHC version:CradleError {cradleErrorDependencies = [], cradleErrorExitCode = ExitFailure 1, cradleErrorStderr = ["Error when calling stack --stack-yaml /home/davo/Documents/IHaskell/./stack-nix.yaml exec --silent ghc -- --numeric-version","",""]}
```

vaibhavsagar:
It seems a bit redundant to add a new `stack.yaml` that's exactly identical to an existing one except for one line. Can we update the instructions to indicate that this setting should be changed in the existing file for easier development?

Issue #1202, opened 2021-01-04
Syntax Highlighting absent in jupyter lab even when jupyterlab-ihaskell installed?


Bodhijesus:
Hi,

I'm on a wsl installation of Ubuntu 18.04, and I made sure to follow all the steps on the page including installing the extension for jupyter lab, and I still get the following result.  To be sure nothing was wrong with the installation itself, I ran a normal jupyter notebook with the Haskell kernel and did get syntax highlighting.  I run jupyter lab simply be calling the `jupyter lab` command.

![image](https://user-images.githubusercontent.com/18375375/103556973-664f4080-4e67-11eb-8e53-f5cceb3e12fc.png)

jamesdbrock:
In September, *IHaskell* was changed so that you no longer need the *jupyterlab-ihaskell* extension to get syntax highlighting. https://github.com/gibiansky/IHaskell/pull/1191

Instead call the `ihaskell` kernel with the option `--codemirror Haskell`.

You can add that option to the `/usr/local/share/jupyter/kernels/haskell/kernel.json` file, (or wherever your `kernel.json` file is stored, run the command `jupyter paths` to see your paths).

Here is an example of a `kernel.json` file for a `stack`-based IHaskell installation:

```
/usr/local/share/jupyter/kernels/haskell $ cat kernel.json 
{"display_name":"Haskell","argv":["/opt/stack/global-project/.stack-work/install/x86_64-linux-tinfo6/fffebb3fddcd52e7f38de3e253571498d665f4223658a7bafc9d5035b02e4ef3/8.8.4/bin/ihaskell","kernel","{connection_file}","--ghclib","/opt/stack/programs/x86_64-linux/ghc-tinfo6-8.8.4/lib/ghc-8.8.4","--codemirror","Haskell","+RTS","-M3g","-N2","-RTS","--stack"],"language":"haskell"}(base)
```

We should change the README instructions for IHaskell.

Bodhijesus:
Thanks for the response!  Unfortunately, an awkward thing I found is that while jupyter lab certainly recognizes the haskell kernel, there is no folder for the kernel in `/usr/local/share/jupyter/kernels`, only python3.  I went ahead and uninstalled the _jupyterlab-ihaskell_ extension _and_ made its own folder inside and created a kernel.json file with those parameters you specified, but with no luck:

[...]

PR #1191, opened 2020-09-22
add codemirror command line flag

MMesch:
How about this? I have tested it with jupyterWith and you can do:

```
  iHaskell = jupyter.kernels.iHaskellWith {
    name = "haskell";
    packages = p: with p; [ p.vinyl p.lens p.singletons p.aeson ];
    extraIHaskellFlags = "--codemirror Haskell";
  };
  jupyterEnvironment =
    jupyter.jupyterlabWith {
      kernels = [ iHaskell ];
    };
```

This gives syntax highlighting without any extension which simplifies the whole setup quite a lot!

Issue #1120, opened 2019-11-18
Syntax Highlighting doesn't work in Lab


AlexeyRaga:
Despite an open issue https://github.com/gibiansky/IHaskell/issues/1017 and a closed issue https://github.com/gibiansky/IHaskell/issues/885 in my case I see tab completion working but no syntax highlighting for the `lab`:

![image](https://user-images.githubusercontent.com/1394391/69046809-f822a100-0a4d-11ea-984c-4f0ec2a6c1e7.png)

I can dedicate some time to make it work if someone directs me to how it should be implemented / fixed.

vaibhavsagar:
Were you able to install the IHaskell JupyterLab extension? IIRC that is what provides the syntax highlighting.

AlexeyRaga:
@vaibhavsagar I am just running `$(nix-build)/bin/ihaskell-lab` hoping that the extension is installed/activated by that script/derivation.

I will try to activate it manually now...

AlexeyRaga:
@vaibhavsagar OK, how do I do that? :)
When I try doing it manyally I get this error:

```
$(nix-build)/bin/jupyter-labextension install ihaskell_labextension
Node v10.16.0

Errored, use --debug for full output:
PermissionError: [Errno 13] Permission denied: '/nix/store/llkjd9yz8v47swmngwj56prvmra74zad-python3.7-jupyterlab-0.35.6/share/jupyter/lab/extensions'
```

Could you please advice how this should be done properly?

vaibhavsagar:
Try using https://github.com/tweag/jupyterWith, which has instructions for including the extension.

AlexeyRaga:
@vaibhavsagar I have managed to install the extension with these steps:
```
$(nix-build)/bin/jupyter-labextension install jupyterlab-ihaskell --app-dir="./jupyterlab" --no-build
$(nix-build)/bin/jupyter-lab build --app-dir="./jupyterlab"
$(nix-build)/bin/jupyter-lab --app-dir="./jupyterlab"
```

It would be nice if `release.nix` would expose a parameter for `--app-dir` so we could do:
```
import "${pkgs.ihaskell}/release.nix" {
  jupyterlabAppDir = "./jupyterlab"
  ...
}
```
Also, iIs there a reason that the extension is not installed by default?

vaibhavsagar:
I'm glad you got it working! I can look into adding a parameter to `release.nix`.

[...]

PR #1223, opened 2021-06-08
Enable CodeMirror syntax highlighting in JupyterLab

### LINKED ISSUE #1199: Enable syntax highlighting in JupyterLab by default
I think Haskell syntax highlighting should be enabled by default. One way to do that would be to insert a switch in `kernel.json`, maybe like

```Haskell
++ ["--codemirror", "Haskell"]
```

after the `--ghclib` argument.

https://github.com/gibiansky/IHaskell/blob/ef698157f44960566687a308e3455b5ba031eb43/src/IHaskell/IPython.hs#L140-L144

But if we did this, then how would syntax highlighting be disabled?

Maybe there's a more natural way to accomplish this.

Reference: https://github.com/gibiansky/IHaskell/pull/1191

jamesdbrock:
Continuation of https://github.com/gibiansky/IHaskell/pull/1191

Resolves #1199

Also will satisfy older issues #1202 #1120

### DIFF
--- a/src/IHaskell/IPython.hs
+++ b/src/IHaskell/IPython.hs
@@ -51,7 +51,7 @@ defaultKernelSpecOptions = KernelSpecOptions
   , kernelSpecRTSOptions = ["-M3g", "-N2"]  -- Memory cap 3 GiB,
                                             -- multithreading on two processors.
   , kernelSpecDebug = False
-  , kernelSpecCodeMirror = "ihaskell"
+  , kernelSpecCodeMirror = "Haskell"
   , kernelSpecConfFile = defaultConfFile
   , kernelSpecInstallPrefix = Nothing
   , kernelSpecUseStack = False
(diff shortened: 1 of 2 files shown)


vaibhavsagar:
We actually define a specific `ihaskell` mode [here]()https://github.com/gibiansky/IHaskell/blob/66333eaf25601d55cfd27760b5bd22866a7526f1/html/kernel.js#L30, so I'm not on board with changing the default. Adding a `--codemirror` sounds good though.

vaibhavsagar:
Is there a way to change the default for JupyterLab only without affecting classic notebooks?

vaibhavsagar:
Alternatively, maybe it is time to change the default and rip out our custom Codemirror mode definition? I'm not sure, I would need to think about it some more. I think making it configurable is a good idea and have no objections there, but I don't want to change the default just yet.

jamesdbrock:
Oh okay, I see what that special CodeMirror "ihaskell" mode is about, yeah this commit probably will not behave correctly.

The CodeMirror built-in mode is name "haskell".

We have a special `multiplexingMode` named "ihaskell" which detects if a code cell is Haskell and should use the "haskell" mode, or if the code cell is a `:` command and should use the "text/plain" mode.

jamesdbrock:
I have some observations to make about the current behavior of `ihaskell` Codemirror on the master branch:

1. `--codemirror haskell` will enable syntax highlighting in Lab.
2. `--codemirror Haskell` will enable syntax highlighting in Lab.
3. `--codemirror ihaskell` will not enable syntax highlighting in Lab.
4. In the classic Notebook no switch is needed, Codemirror is already enabled.

PR #1225, opened 2021-06-09
hie.yaml: init

vaibhavsagar:


vaibhavsagar:
@jamesdbrock, @daviddavo, I got the included `hie.yaml` to work with Nix and Cabal but it's a little involved:

1. Make sure you're using Nixpkgs 21.05:
```bash
$ nix-channel --add https://nixos.org/channels/nixos-21.05 nixpkgs
$ nix-channel --update
```

2. Install `cabal-install` and `haskell-language-server` globally:
```bash
$ nix-env -i cabal-install
$ nix-env -i haskell-language-server
```
3. Remove all the `ihaskell-*` packages from `cabal.project` except `ihaskell-widgets`, this is what it should look like afterwards:
```
packages: .
          ./ipython-kernel
          ./ghc-parser
          ./ihaskell-display/ihaskell-widgets
```

4. Open VS Code **from inside `nix-shell`**:
```bash
$ nix-shell release-8.10.nix -A passthru.haskellPackages.ihaskell.env
$ code .
```

5. You should be able to navigate to a `ihaskell-widgets` source file and watch it build `singletons` from source, after which it will give you useful suggestions.

jamesdbrock:
This works for me for **ihaskell**, **ghc-parser**, and **ipython-kernel**, and HLS can run for **ihaskell-display/ihaskell-widgets**, but right now HLS is failing to build **ihaskell-widgets** with errors like `Could not find module ‘Data.Singletons.TH’`...

jamesdbrock:
I think what we really need is a shell dev environment in which `IHaskell $ cabal v2-build` succeeds. If we had that, then I'm pretty sure HLS would work with no other setup.

vaibhavsagar:
That's impossible with just Nix, since `cabal` has its own setup that cannot be automated. In this case it looks like it might be using the newest version of `singletons` which splits out that module into `singletons-th`. We might need to use https://www.stackage.org/lts-17.15/cabal.config to make sure we're using the correct versions. I will try that and update the instructions.

vaibhavsagar:
@jamesdbrock that seemed to help. I've updated the instructions.

Issue #870, opened 2018-06-04
ihaskell-widgets runs with no error, however, displays no widget


eldad-a:
Hello,

The short version:
ihaskell-widgets runs with no error, however shows no widget.

Long version:
This is my first experience with `Haskell` and I am thankful to the Devs for creating an interface based on `jupyter notebook`, which I find very friendly.

The following example code executes with no error, however, with no widget displayed either.

```haskell
import IHaskell.Display.Widgets
slider <- mkFloatSlider

setField slider ChangeHandler $ do
    value <- getField slider FloatValue
    print value
    
slider
```
The above is copied from [Demo of IHaskell Notebook](https://begriffs.com/posts/2016-01-20-ihaskell-notebook.html) by [Andew Giabiansky](http://andrew.gibiansky.com/pages/about.html), January 20, 2016

To run the example notebooks from the `ihaskell-widgets` [IHaskell repository](https://github.com/gibiansky/IHaskell) I had to rename the `stack.yaml` in the `ihaskell-display/ihaskell-widgets/`, as suggested in [Cannot install ihaskell-display widgets with stack #715
](https://github.com/gibiansky/IHaskell/issues/715) .

Then the examples (two notebooks I tried) run with no errors, but still, no widget is displayed.

Hope you could guide me how to resolve this matter.

vaibhavsagar:
It looks like `ihaskell-widgets` hasn't been updated in some time (beyond version bumps by me) and the protocol version it speaks is out of date. I'm not sure when I'll get around to fixing this and I'm sorry I don't have a better answer for you!

![ihaskell-widget-protocol-mismatch](https://user-images.githubusercontent.com/1525767/40951723-c51159a0-68aa-11e8-9f12-2b245f531519.png)

vaibhavsagar:
This issue looks related: https://github.com/JuliaGizmos/Interact.jl/issues/176

vaibhavsagar:
For myself or whoever decides to implement this: here is [a document describing the new protocol](https://github.com/jupyter-widgets/ipywidgets/blob/3023b3ed24fbb172eb44643798a603aaed15ae76/packages/schema/messages.md) and the changes would probably have to made to [`IHaskell.Display.Widgets`](https://github.com/gibiansky/IHaskell/blob/00c762b77b36b08bb7baf4a3b84f251bd9b6be90/src/IHaskell/Eval/Widgets.hs).

eldad-a:
@vaibhavsagar 

[...]

PR #1242, opened 2021-07-30
Updating IHaskell-widgets

daviddavo:
This pull request aims to solve the issue #870, updating the new widget protocol and implementing any new widgets

Things to do:
- [ ] Update sample notebooks
- [x] Implement the Link widgets
- [x] Implement jslink and jsdlink as helper functions
- [ ] Implement the Controller widget
- [x] Revise and improve documentation
- [x] Fix the OutputWidget


ihaskell-display/ihaskell-widgets/ihaskell-widgets.cabal
```diff
@@ -110,10 +138,7 @@ library
 
 					 -- The singletons package version is locked to the compiler
 					 -- so let cabal choose the right one.
-                     , singletons -any
-
-  if impl (ghc >= 9.0)
-    build-depends:     singletons-base -any
+                     , singletons < 3.0.0
```
vaibhavsagar, line 141: I think this section didn't rebase correctly, can you add what we have in `master` so it works on GHC 9.0?

ihaskell-display/ihaskell-widgets/src/IHaskell/Display/Widgets/Box/HBox.hs
```diff
@@ -0,0 +1,51 @@
...
+{-# OPTIONS_GHC -fno-warn-orphans  #-}
+
+module IHaskell.Display.Widgets.Box.HBox
+  ( -- * The HBox widget
+    HBox
+    -- * Constructor
+  , mkHBox
+  ) where
+
+-- To keep `cabal repl` happy when running from the ihaskell repo
```
vaibhavsagar, line 15: Can we remove this?

vaibhavsagar: Hmm, I see that this same line is present in the codebase prior to this change. I guess it doesn't really do any harm, It's up to you if you want to leave it in or remove it everywhere else too.

daviddavo: Should I remove also the `mport Prelude`? Or only the comment?

vaibhavsagar: The `import Prelude` as well, but given how it's all over the `ihaskell-widgets` code it doesn't particularly matter.

daviddavo:
There is a problem with capturing the output using the output widget. The output widget is able to capture the output if you set the `MsgID` attribute to the parent message id of messages to capture. This is thanks to the frontend, so you don't have problems with capturing stdout/stderr. So we could do function like:

```haskell
captureOutput :: OutputWidget -> IO a -> IO ()
captureOutput o f = do
  setField o MsgID (getSomehowTheMsgID)
  f
  setField o MsgID ""
  return ()
```

The problem is, how could we get the id of the `execution_request`?

.github/workflows/docker.yml
```diff
@@ -1,5 +1,7 @@
 name: "Docker"
 on:
+  workflow_dispatch:
```
vaibhavsagar, line 3: Why are you updating the Docker workflow?

.github/workflows/docker.yml
```diff
@@ -23,7 +25,7 @@ jobs:
     - uses: actions/checkout@01aecccf739ca6ff86c0539fbc67a7a5007bbc81
     - uses: elgohr/Publish-Docker-Github-Action@f7aca2fea76a5218f3c76cd5933c3ba1d8008774
       with:
-        name: gibiansky/ihaskell
+        name: daviddavo/ihaskell
```
vaibhavsagar, line 28: I don't think this belongs in this PR, I would suggest making these changes on a separate branch.

Widget List.ipynb
```diff
@@ -0,0 +1,1495 @@
+{
```
vaibhavsagar, line 1: Please add any extra notebooks to the `notebooks` directory.

src/IHaskell/Types.hs
```diff
@@ -277,6 +295,12 @@ dupHeader hdr messageType = do
   uuid <- liftIO random
   return hdr { mhMessageId = uuid, mhMsgType = messageType }
 
+-- | Modyfies a header and appends the version of the Widget Messaging Protocol as metadata
```
vaibhavsagar, line 298:
```suggestion
-- | Modifies a header and appends the version of the Widget Messaging Protocol as metadata
```

src/IHaskell/Types.hs
```diff
@@ -41,8 +43,10 @@ module IHaskell.Types (
 
 import           IHaskellPrelude
 
-import           Data.Aeson (ToJSON (..), Value, (.=), object)
+import qualified Data.HashMap.Strict as HashMap
```
vaibhavsagar, line 46: I don't think we need these additional imports if we use `object`.

Widget List.ipynb
daviddavo: This file was supposed to be ignored as is from IPython...

The widgets notebooks are on ihaskell-display/Examples

src/IHaskell/Types.hs
```diff
@@ -277,6 +295,12 @@ dupHeader hdr messageType = do
   uuid <- liftIO random
   return hdr { mhMessageId = uuid, mhMsgType = messageType }
 
+-- | Modyfies a header and appends the version of the Widget Messaging Protocol as metadata
+setVersion :: MessageHeader  -- ^ The header to modify
+           -> String         -- ^ The version to set
+           -> MessageHeader  -- ^ The modified header
+setVersion hdr v = hdr { mhMetadata = Metadata (HashMap.fromList [("version", String $ pack v)]) }
```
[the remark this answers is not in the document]
daviddavo: The problem is that `Metadata` is `newtype Metadata = Metadata Object`, but the [`object`](https://hackage.haskell.org/package/aeson-1.5.6.0/docs/Data-Aeson.html#v:object) function returns a `Value`.

I saw that they use through the docs the `fromList` function to create an `Object`.

Another alternative could be extracting the `Object` with a lambda expression like `(\(Object o) -> o) $ object ["version .= pack v]`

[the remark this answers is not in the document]
vaibhavsagar: Oh, good point, I didn't realise that. The way you currently do it is fine.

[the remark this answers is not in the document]
daviddavo: I'll add a comment in case we forget about this later

ihaskell-display/ihaskell-widgets/README.md
```diff
@@ -5,3 +5,81 @@ IHaskell. The frontend (javascript) is provided by the jupyter/ipython notebook
...
+- [ ] Overload setField so it can be used with `Maybes` or other wrapper types without having to put `Just` every time.
+- [ ] Add some "utils" work:
+    - [ ] Create media widget from file
+    - [ ] Get the selected label from a selection value
+
+## How to...
+This is a mini-guide for developers that want to update to a more recent widgets specification, but without
+dwelling into the deeps of the project
+
+### Add a new attribute
```
jamesdbrock, line 33: These instructions are really nice.

src/IHaskell/Types.hs
vaibhavsagar: Turns out we can't use `object`.

src/IHaskell/Eval/Widgets.hs
```diff
@@ -18,14 +18,27 @@ import           Control.Concurrent.STM (atomically)
...
+
+import           Data.Foldable (foldl)
 import           System.IO.Unsafe (unsafePerformIO)
 
 import           IHaskell.Display
 import           IHaskell.Eval.Util (unfoldM)
 import           IHaskell.IPython.Types (showMessageType)
 import           IHaskell.Types
 
+#if MIN_VERSION_base(4,11,0)
```
vaibhavsagar, line 34: Nit: I'd prefer if we removed this and used plain `fmap` or `<$>` instead.

vaibhavsagar (approved):
Looks good to me!

src/IHaskell/Eval/Widgets.hs
vaibhavsagar: Resolved!

jamesdbrock (approved):

PR #1246, opened 2021-08-15
Updated sample widget Notebooks

daviddavo:
I've updated all the notebooks to work with the new ihaskell-widgets version.

I also added ChangeHandlers to the datepicker and color widgets.


vaibhavsagar:
Thanks!

PR #1247, opened 2021-08-15
replaceOutput runs without flickering the GUI

jamesdbrock:
Also export `OutputMsg` constructors so that users can call
`setField` on an `OutputWidget` to write their own `replaceOutput`
functions.

### DIFF
--- a/ihaskell-display/ihaskell-widgets/src/IHaskell/Display/Widgets.hs
+++ b/ihaskell-display/ihaskell-widgets/src/IHaskell/Display/Widgets.hs
@@ -77,4 +77,6 @@ import           IHaskell.Display.Widgets.Common as X
 import           IHaskell.Display.Widgets.Types as X (setField, getField, properties, triggerDisplay,
                                                       triggerChange, triggerClick, triggerSelection,
                                                       triggerSubmit, ChildWidget(..), StyleWidget(..),
-                                                      WidgetFieldPair(..), Date(..), unlink, JSONByteString(..))
+                                                      WidgetFieldPair(..), Date(..), unlink,
+                                                      JSONByteString(..), OutputMsg(..))
+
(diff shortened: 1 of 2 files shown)


jamesdbrock:
So, this seems to work and solve the problem of the GUI flickering when `replaceOutput` is called.

Question: The old `replaceOutput` called

```haskell
_ <- setField w MsgID ""
```

but the new one does not. Was that `setField w MsgID ""` important for some reason?

daviddavo:
The MsgID attribute doesn't change, it should always be set to `""`, that's why I removed that line...

But setting the field value makes it send an update message, maybe it needs that message...

vaibhavsagar:
These changes conflict with https://github.com/gibiansky/IHaskell/pull/1246, does still need need to be merged?

ihaskell-display/ihaskell-widgets/src/IHaskell/Display/Widgets.hs
```diff
@@ -77,4 +77,6 @@ import           IHaskell.Display.Widgets.Common as X
 import           IHaskell.Display.Widgets.Types as X (setField, getField, properties, triggerDisplay,
                                                       triggerChange, triggerClick, triggerSelection,
                                                       triggerSubmit, ChildWidget(..), StyleWidget(..),
-                                                      WidgetFieldPair(..), Date(..), unlink)
+                                                      WidgetFieldPair(..), Date(..), unlink,
+                                                      JSONByteString(..), OutputMsg(..))
```
daviddavo, line 81: The conflict comes because I forgot to add the JSONByteString constructor to the export list. It's here, though, so it should be safe to just accept this file and ignore mine.

vaibhavsagar: @jamesdbrock can you rebase?

jamesdbrock: Rebased.

vaibhavsagar:
Thanks!

PR #1249, opened 2021-08-16
Update some example notebooks

jamesdbrock:
Update some example notebooks to the latest notebook version, so that
when they start, Jupyter doesn't immediately update them and mark them
as changed.

Also replace the fake diagrams test notebook with a real diagrams test
notebook. The fake diagrams test notebook was actually the charts test
notebook.


vaibhavsagar:
Thanks!

Issue #1251, opened 2021-08-16
Docker image fails with "Bad interface file"


brandon-leapyear:
Repro:
* `docker run --rm -p 8888:8888 gibiansky/ihaskell`
* Open `localhost:8888`
* Create a new notebook in the top-level directory
* Type in `putStrLn "Hello world!"`, shift + Enter
* Get error:
```
ihaskell: Bad interface file: /home/jovyan/.stack/programs/x86_64-linux/ghc-8.8.4/lib/ghc-8.8.4/base-4.13.0.0/Prelude.hi
    magic number mismatch: old/corrupt interface file? (wanted 33214052, got 1)
Bad interface file: /home/jovyan/.stack/programs/x86_64-linux/ghc-8.8.4/lib/ghc-8.8.4/directory-1.3.6.0/System/Directory.hi
    magic number mismatch: old/corrupt interface file? (wanted 33214052, got 1)
Bad interface file: /home/jovyan/.stack/programs/x86_64-linux/ghc-8.8.4/lib/ghc-8.8.4/unix-2.7.2.2/System/Posix/IO.hi
    magic number mismatch: old/corrupt interface file? (wanted 33214052, got 1)
Bad interface file: /home/jovyan/.stack/programs/x86_64-linux/ghc-8.8.4/lib/ghc-8.8.4/base-4.13.0.0/System/IO.hi
    magic number mismatch: old/corrupt interface file? (wanted 33214052, got 1)
Bad interface file: /home/jovyan/.stack/programs/x86_64-linux/ghc-8.8.4/lib/ghc-8.8.4/template-haskell-2.15.0.0/Language/Haskell/TH.hi
    magic number mismatch: old/corrupt interface file? (wanted 33214052, got 1)
```

My system:
* Mac 10.15.7
* Docker version 20.10.7, build f0df350

brandon-leapyear:
It works after manually ssh-ing into the Docker container and editing `~/.stack/global-project/stack.yaml` to use the same resolver as `/home/jovyan/ihaskell/stack.yaml`.

The instructions should probably include something like "make sure to create new notebooks within a directory containing a `stack.yaml` with the same resolver as the `stack.yaml` in `/home/jovyan/ihaskell/`". Right now, the instructions only talk about "check the resolver is the same" when mounting volumes

brandon-leapyear:
My current workaround script: https://github.com/brandon-leapyear/config/blob/main/bin/ihaskell

[...]

PR #1252, opened 2021-08-18
Rewrite Docker image

brandon-leapyear:
Reimplements `Dockerfile` with the following changes:
* Uses the [`haskell` Docker image](https://hub.docker.com/_/haskell) as a base image, so we don't have to worry about installing Stack
* Uses multistage builds, to avoid bulking the final image with compile-time artifacts.
    * Currently, the `gibiansky/ihaskell:202108170429266e69c2` image is 6.26 GB on my computer, after a `docker pull`
    * This PR builds a docker image that's 1.93 GB. Note: the `haskell:8.10.4` base image we're using is 1.52 GB, so this PR builds a docker image that only adds 400 MB to the base image!
    * This should greatly improve download times
* Writes `~/.stack/global-project/stack.yaml` to have the same resolver as the in-repo `stack.yaml` file (resolves https://github.com/gibiansky/IHaskell/issues/1251#issuecomment-899985184)
* Adds a `--env-file` flag, to fix using `stack` from the `haskell` base image (the `stack exec env` command called [here](https://github.com/gibiansky/IHaskell/blob/6e69c210c5127d74870a8d3645fc492a7f064eba/main/Main.hs#L131) was reinstalling GHC; didn't seem to be an easy way to tell it to use `--system-ghc`)

TODO:
- [ ] Fix syntax highlighting
    - Syntax highlighting is no longer working. I'm looking at the old version of the `Dockerfile`, and I'm not sure how it does syntax highlighting, since it's not `COPY`ing the `jupyterlab-ihaskell` directory. @vaibhavsagar @gibiansky @gregwebs  do you know how the current Docker image gets syntax highlighting? (Possibly related: https://github.com/gibiansky/IHaskell/issues/1199)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,6 @@ on:
     paths-ignore:
       - 'demo/**'
       - 'docker/**'
-      - 'Dockerfile'
       - '.dockerignore'
       - '.ghci'
       - '.gitignore'
@@ -20,7 +19,6 @@ on:
     paths-ignore:
       - 'demo/**'
       - 'docker/**'
-      - 'Dockerfile'
       - '.dockerignore'
       - '.ghci'
       - '.gitignore'
@@ -94,3 +92,16 @@ jobs:
           stack install
           stack exec -- ihaskell install --stack
           test/acceptance.nbconvert.sh stack exec -- jupyter nbconvert
+      - name: Check Dockerfile GHC version matches
+        run: |
+          if [[ '${{ matrix.versions.stack-yaml }}' != 'stack.yaml' ]]; then
+            exit 0
+          fi
+
+          set -e
+          STACK_GHC_VERSION=$(stack exec -- ghc --version | awk '{ print $NF }')
+          DOCKER_GHC_VERSION=$(sed -n 's/ARG GHC_VERSION=\(.*\)/\1/p' Dockerfile)
+          if [[ ${STACK_GHC_VERSION} != ${DOCKER_GHC_VERSION} ]]; then
+            echo 'GHC_VERSION in Dockerfile does not match stack resolver'
+            exit 1
+          fi
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,7 @@ on:
     paths-ignore:
       - 'demo/**'
       - 'docker/**'
+      - 'Dockerfile'
       - '.dockerignore'
       - '.ghci'
       - '.gitignore'
@@ -19,6 +20,7 @@ on:
     paths-ignore:
       - 'demo/**'
       - 'docker/**'
+      - 'Dockerfile'
       - '.dockerignore'
       - '.ghci'
       - '.gitignore'
--- a/Dockerfile
+++ b/Dockerfile
@@ -88,11 +88,14 @@ WORKDIR ${HOME}
 COPY --from=builder --chown=${NB_UID} /build/resolver.txt /tmp/
 RUN stack setup --resolver=$(cat /tmp/resolver.txt) --system-ghc
 
+# Set up env file
+RUN stack exec env --system-ghc > ${IHASKELL_DATA_DIR}/env
+
 # Install + setup IHaskell
 COPY --from=builder --chown=${NB_UID} /build/bin/ihaskell /usr/local/bin/
 COPY --from=builder --chown=${NB_UID} /build/html ${IHASKELL_DATA_DIR}/html
 RUN export ihaskell_datadir=${IHASKELL_DATA_DIR} && \
-    ihaskell install --stack
+    ihaskell install --env-file ${IHASKELL_DATA_DIR}/env
 RUN jupyter notebook --generate-config
 
 CMD ["jupyter", "notebook", "--ip", "0.0.0.0"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,58 +1,98 @@
-FROM ubuntu:18.04
+# should match the GHC version of the stack.yaml resolver
+FROM haskell:8.10.4 AS ihaskell_base
 
-ARG STACK_VERSION=2.7.3
-ARG RESOLVER=lts-17.4
-
-# Install all necessary Ubuntu packages
-RUN apt-get update && apt-get install -y python3-pip libgmp-dev libmagic-dev libtinfo-dev libzmq3-dev libcairo2-dev libpango1.0-dev libblas-dev liblapack-dev gcc g++ wget && \
+# Install Ubuntu packages needed for IHaskell runtime
+RUN apt-get update && \
+    apt-get install -y \
+            libblas3 \
+            libcairo2 \
+            liblapack3 \
+            libmagic1 \
+            libpango-1.0-0 \
+            libzmq5 \
+        && \
     rm -rf /var/lib/apt/lists/*
 
+FROM ihaskell_base AS builder
+
+# Install Ubuntu packages needed for IHaskell build
+RUN apt-get update && \
+    apt-get install -y \
+            libblas-dev \
+            libcairo2-dev \
+            liblapack-dev \
+            libmagic-dev \
+            libpango1.0-dev \
+            libzmq3-dev \
+        && \
+    rm -rf /var/lib/apt/lists/*
+
+WORKDIR /build
+
+# Build snapshot
+COPY stack.yaml stack.yaml
+COPY ihaskell.cabal ihaskell.cabal
+COPY ipython-kernel ipython-kernel
+COPY ghc-parser ghc-parser
+COPY ihaskell-display ihaskell-display
+RUN stack setup
+RUN stack build --only-snapshot
+
+# Build IHaskell itself.
+# Don't just `COPY .` so that changes in e.g. README.md don't trigger rebuild.
+COPY src src
+COPY html html
+COPY main main
+COPY LICENSE LICENSE
+RUN stack install --local-bin-path ./bin/
+
+# Save resolver used to build IHaskell
+RUN sed -n 's/resolver: \(.*\)/\1/p' stack.yaml | tee resolver.txt
+
+# Save third-party data files
+RUN mkdir /data && \
+    snapshot_install_root=$(stack path --snapshot-install-root) && \
+    cp $(find ${snapshot_install_root} -name hlint.yaml) /data
+
+FROM ihaskell_base AS ihaskell
+
 # Install Jupyter notebook
+RUN apt-get update && \
+    apt-get install -y python3-pip && \
+    rm -rf /var/lib/apt/lists/*
 RUN pip3 install -U jupyter
 
-ENV LANG C.UTF-8
-ENV LC_ALL C.UTF-8
+# Create runtime user
 ENV NB_USER jovyan
 ENV NB_UID 1000
-ENV HOME /home/${NB_USER}
-
 RUN adduser --disabled-password \
     --gecos "Default user" \
     --uid ${NB_UID} \
     ${NB_USER}
 
-RUN wget -qO- https://github.com/commercialhaskell/stack/releases/download/v$STACK_VERSION/stack-$STACK_VERSION-linux-x86_64.tar.gz | tar xz --wildcards --strip-components=1 -C /usr/bin '*/stack'
+# Create directory for storing ihaskell files
+ENV IHASKELL_DATA_DIR /usr/local/lib/ihaskell
+RUN mkdir -p ${IHASKELL_DATA_DIR} && chown ${NB_UID} ${IHASKELL_DATA_DIR}
 
-# Set up a working directory for IHaskell
-RUN install -d -o ${NB_UID} -g ${NB_UID} ${HOME} ${HOME}/ihaskell
-WORKDIR ${HOME}/ihaskell
+# Set up + set hlint data directory
+ENV HLINT_DATA_DIR /usr/local/lib/hlint
+COPY --from=builder --chown=${NB_UID} /data/hlint.yaml ${HLINT_DATA_DIR}/
+ENV hlint_datadir ${HLINT_DATA_DIR}
 
+# Set current user + directory
 USER ${NB_UID}
-
-# Install dependencies for IHaskell
-COPY --chown=${NB_UID}:${NB_UID} stack.yaml stack.yaml
-COPY --chown=${NB_UID}:${NB_UID} ihaskell.cabal ihaskell.cabal
-COPY --chown=${NB_UID}:${NB_UID} ipython-kernel ipython-kernel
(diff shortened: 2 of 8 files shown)


vaibhavsagar:
Can you separate your changes to the resolver we use from the Dockerfile update? I don't think they both belong in the same PR.

vaibhavsagar:
To answer your question about JupyterLab syntax highlighting, I don't think that's enabled on the current Docker image either so don't worry about supporting that in your PR.

Dockerfile
```diff
@@ -1,58 +1,101 @@
-FROM ubuntu:18.04
+# should match the GHC version of the stack.yaml resolver
+FROM haskell:8.10.4 AS ihaskell_base
```
vaibhavsagar, line 2: This still needs to be manually updated, right? Or am I missing something?

brandon-leapyear:
> To answer your question about JupyterLab syntax highlighting, I don't think that's enabled on the current Docker image either so don't worry about supporting that in your PR.

That's not true. I can run `docker run --rm -p 8888:8888 gibiansky/ihaskell:202108170429266e69c2` today, and create a new notebook, and it'll have syntax highlighting

Dockerfile
brandon-leapyear: oh right sure. i can make that automatic too; it would just make the `docker build` command a bit more complicated. lemme update the PR

brandon-leapyear: How are you currently continuously-deploying the docker image? Is it via CI or Docker Hub's build feature?

I usually solve this problem by adding an `ARG` and writing a wrapper script that reads `stack.yaml` and sets the `--build-arg` automatically, but I don't think that would work if you deploy with Docker Hub. If you use Docker Hub, what we can do instead is keep this hardcoded, then add a test in CI to check that it matches `stack.yaml`. So at least there'd be an alert

vaibhavsagar:
Yes, that's the classic Jupyter interface, not JupyterLab. We implement syntax highlighting by copying [`kernel.js`](https://github.com/gibiansky/IHaskell/blob/6e69c210c5127d74870a8d3645fc492a7f064eba/html/kernel.js): https://github.com/gibiansky/IHaskell/blob/0f1262d3e710518fd734fbda6f2eba33e476836b/src/IHaskell/IPython.hs#L160-L169

Dockerfile
vaibhavsagar: I'm using CI to update the Docker image every time `master` is updated: [docker.yml](https://github.com/gibiansky/IHaskell/blob/6e69c210c5127d74870a8d3645fc492a7f064eba/.github/workflows/docker.yml)

vaibhavsagar:
One other thing I should mention before you sink more time into this: the resulting `Dockerfile` needs to be buildable by mybinder.org like [the current one is](https://mybinder.org/v2/gh/gibiansky/IHaskell/master), if you come up with a fancy multi-stage build that does not have this property I won't be able to accept your PR.

brandon-leapyear:
> Yes, that's the classic Jupyter interface, not JupyterLab. We implement syntax highlighting by copying [`kernel.js`](https://github.com/gibiansky/IHaskell/blob/6e69c210c5127d74870a8d3645fc492a7f064eba/html/kernel.js):

I still have `kernel.js` copied, so syntax highlighting should work, right? Why might it not work?

ihaskell.cabal
```diff
@@ -127,7 +127,7 @@ executable ihaskell
   hs-source-dirs: main
   other-modules:
                    Paths_ihaskell
-  ghc-options: -threaded -rtsopts -Wall -dynamic
+  ghc-options: -threaded -rtsopts -Wall
```
vaibhavsagar, line 130: Why was this flag removed?

brandon-leapyear: IMO this flag shouldn't have been added in the first place. If someone needs this flag, they can add it with `--ghc-options -dynamic`.

It simplifies the Docker image by building a statically linked binary, so that I don't have to copy all the `.so` files into the runtime image. There's probably a `-no-dynamic` flag that I could use to disable it, but, again, executables shouldn't be built dynamically by default

brandon-leapyear:
> > Yes, that's the classic Jupyter interface, not JupyterLab. We implement syntax highlighting by copying [`kernel.js`](https://github.com/gibiansky/IHaskell/blob/6e69c210c5127d74870a8d3645fc492a7f064eba/html/kernel.js):
> 
> I still have `kernel.js` copied, so syntax highlighting should work, right? Why might it not work?

oh never mind... it seems like syntax highlighting turns on after reloading the page... Specifically, it seems like the notebook initially requests `/static/components/codemirror/mode/ihaskell/ihaskell.js` and gets a 404, then reloading fetches `/static/components/codemirror/mode/haskell/haskell.js` (note: `ihaskell` => `haskell`) and gets a 200.

ihaskell.cabal
vaibhavsagar: I added it for a reason: https://gitlab.haskell.org/ghc/ghc/-/issues/15524

brandon-leapyear: hm the PR that added this flag doesn't mention that issue at all: https://github.com/gibiansky/IHaskell/pull/1184

Seems to only affect the Nix build, is that right? I notice the Nix builds in this CI increased from ~1min to ~3min, but the Linux build takes roughly the same amount of time

vaibhavsagar: Did you read the linked issue? The problem isn't Nix-specific.

brandon-leapyear: I did read the issue, but I said

> I notice the Nix builds in this CI increased from ~1min to ~3min, but the Linux build takes roughly the same amount of time.

So I don't see any performance regression in the non-Nix side. I also don't see any obvious performance issues using the Docker image built from this branch

vaibhavsagar: Okay, I tried it myself without `-dynamic` and the performance on Nix was fine, I didn't notice any issues either.

brandon-leapyear:
FYI I wrote a script that runs `ihaskell` with Docker (optionally mounting the current directory or a specified directory) and automatically opens the browser. Not sure if this should be added somewhere to the repo/README:

https://github.com/brandonchinn178/config/blob/main/bin/ihaskell

vaibhavsagar:
Feel free to add a link to it in the README under the "Docker" section.

gregwebs:
sorry, I don't think I tested syntax highlighting.

.github/workflows/ci.yml
```diff
@@ -20,7 +19,6 @@ on:
     paths-ignore:
       - 'demo/**'
       - 'docker/**'
-      - 'Dockerfile'
```
vaibhavsagar, line 23: Can you change these lines back? I don't want the CI to run when only changes to the Dockerfile are made. When we update the Dockerfile, it's extremely unlikely that we'll update it to be out of date with the `stack.yaml`, so when we make a change to `stack.yaml` that is likely to move things out of sync, the lines you've added below to check the Dockerfile version will still run and notify us appropriately.

.github/workflows/ci.yml
```diff
@@ -94,3 +92,16 @@ jobs:
           stack install
           stack exec -- ihaskell install --stack
           test/acceptance.nbconvert.sh stack exec -- jupyter nbconvert
+      - name: Check Dockerfile GHC version matches
```
vaibhavsagar, line 95: I have to say it's pretty funny that the solution you've ended up with still has the same flaw w.r.t multiple sources of truth that our current setup does. We could add these lines alone and get 90% of the benefit of this PR.

.github/workflows/ci.yml
brandon-leapyear: sure

.github/workflows/ci.yml
brandon-leapyear: Yeah, that's true, although I disagree about the 90% benefit. I would say the biggest improvement this PR makes is bringing the image size down from 6 GB to 1.5 GB.

brandon-leapyear: And actually, the original multiple-sources-of-truth dealt with the `~/.stack/global-project/stack.yaml` resolver and the in-repo `stack.yaml` resolver. That is now one source of truth here: https://github.com/gibiansky/IHaskell/pull/1252/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R91

The `GHC_VERSION` in the `Dockerfile` is just the GHC preloaded into the Docker image. So actually, if `stack.yaml` resolver used a different version of GHC, it would just download another GHC into the build image, but the runtime image's global stack resolver would still match the in-repo stack resolver.

vaibhavsagar: I don't really care about that, since an IHaskell Docker image built with Nix is also quite small: https://vaibhavsagar.com/blog/2019/08/11/ihaskell-nix-docker/.

vaibhavsagar: Actually the Nix-built size comparison isn't fair, I just imported the archive and the resulting image is nearly 4GB so your Dockerfile is in fact the one that results in an image that is more than twice as small.

vaibhavsagar: If by chance you want to see for yourself, the Nix expression is here: https://gist.github.com/vaibhavsagar/1cf8ca927a8ee872d43619c4ac2d2665.

.github/workflows/ci.yml
vaibhavsagar: Can you remove it here as well, please? For the same reason as before, I don't want to run CI when only the Dockerfile has changed since our CI doesn't build the Docker image.

.github/workflows/ci.yml
brandon-leapyear: > Also, you've moved the problem at the expense of additional complexity in the Dockerfile and the IHaskell source code itself that I would have to maintain and update in the presence of changes, so it's arguable whether this is an improvement.

Moving the problem is separate from the additional complexity I added to the Dockerfile/source code. I could've solved this problem by making a PR with just
```diff
@@ line 4 @@
-ARG RESOLVER=lts-17.4

@@ line 49 @@
 RUN stack build && stack install
 RUN mkdir -p ${HOME}/.stack/global-project && \
+    RESOLVER=$(sed -n 's/resolver: \(.*\)/\1/p' stack.yaml) && \
     echo "packages: []\nresolver: ${RESOLVER}\n" > ${HOME}/.stack/global-project/stack.yaml
```
All the complexity in this PR is due to me bringing Dockerfile best practices to the Dockerfile to produce a smaller final image using the `haskell` base docker image (which allowed us to avoid manually installing `stack`)

> It's never explicitly been our focus to provide a production-quality Docker image.

That's fair, but I would also ask why it's not. As a user, it's a bit difficult to quickly install + get set up with IHaskell if I need to install a bunch of dependencies and build it from scratch with Stack. If there were pre-built binaries that one could download, that would help, but since IHaskell links a bunch of C libraries (e.g. cairo), one would still need to install dependencies, and I personally try to avoid globally installing extra dependencies.

I might be open to taking ownership of maintaining the `Dockerfile` in this repo, if that's your primary concern.

vaibhavsagar:
I'm trying to run this locally, and it appears to be failing at this step, is there something I'm doing wrong?

```
Step 30/41 : COPY --from=builder --chown=${NB_UID} /data/hlint.yaml ${HLINT_DATA_DIR}/
unable to convert uid/gid chown string to host mapping: can't find uid for user ${NB_UID}: no such user: ${NB_UID}
```

brandon-leapyear:
What version of Docker are you using? It seems like using env vars in `--chown` was added in 19.03.0: https://github.com/moby/moby/issues/35018#issuecomment-535188969

vaibhavsagar:
Ah, mine is too old (`Docker version 18.09.7, build 2d0083d`), that must be why. I'll try with a newer version, just want to confirm that everything works correctly since we're not building it as part of CI yet.

vaibhavsagar:
It builds locally for me, but not on mybinder.org: https://mybinder.org/v2/gh/brandon-leapyear/IHaskell/chinn/docker. It looks like the mybinder.org builder doesn't have enough memory perhaps?

vaibhavsagar:
I think you've addressed all my feedback, is this ready to merge or are there more changes you'd like to make?

brandon-leapyear:
> It builds locally for me, but not on mybinder.org: https://mybinder.org/v2/gh/brandon-leapyear/IHaskell/chinn/docker. It looks like the mybinder.org builder doesn't have enough memory perhaps?
> 
> Edit: Oh never mind, this isn't a blocker because I've just discovered the existing Dockerfile has the same behaviour.

Yeah, it happens already anyway. We can fix it by adding `--jobs 1` to the `stack build` steps, possibly hidden behind a `--build-arg` that we can override in the continuous deploy job.



> I think you've addressed all my feedback, is this ready to merge or are there more changes you'd like to make?

Nope, I'm good if you are!

vaibhavsagar:
Thanks so much!

ihaskell.cabal
junjihashimoto: @brandon-leapyear @vaibhavsagar 
Hi, in my use case like the following link, when the `-dynamic` is deleted, ihaskell does not load some packages.
https://github.com/hasktorch/hasktorch-jupyter/blob/master/Dockerfile

ihaskell needs to load packages at runtime and does not work entirely with a single binary.
It certainly works with ghc-options -dynamic, but I had to rebuild again this time.

junjihashimoto: I think anything using FFI is affected as there is a problem loading c/c++ DLLs.

junjihashimoto: This option is needed to link c/c++ system libraries not haskell libraries.

brandon-leapyear: What do you mean by "I had to rebuild again this time"? It seems like you're [installing `ihaskell` from source](https://github.com/hasktorch/hasktorch-jupyter/blob/42dfb4ab7c213a3b91f2a105126d39857d51ff47/Dockerfile#L59-L67) already, so it's not like you're rebuilding an artifact that was already built by this Docker image.

If you mean that you built the Docker image once, realized it didn't work, and then have to rebuild the Docker image with `--ghc-options -dynamic`, that's just normal Docker development. If you're writing a `Dockerfile` and you used the wrong arguments in a command, you just have to go back and rebuild the image with the correct arguments.

Unless, are you saying that `ihaskell` depends on something that needs to be loaded dynamically? I haven't been running into this, it must be specific to your project. If it's something specific to your project, you should add `--ghc-options -dynamic` yourself; you shouldn't force everyone building `ihaskell` to use `-dynamic` just because you need it

junjihashimoto: In other words, without this option, ihaskell will not be able to link to libraries that ihaskell does not directly link.
When ihaskell is built, it doesn't know what haskell packages are loaded, and it don't check what c/c++ libraries those haskell packages link to.

PR #1254, opened 2021-08-21
stack.yaml: bump resolver to lts-18.7

vaibhavsagar:


vaibhavsagar:
I am aware that the GHC version in the Dockerfile also needs to be updated but I think this is a good opportunity for us to see if we get a build failure like we are supposed to.

PR #1267, opened 2021-08-28
jupyterlab-ihaskell: make compatible with JupyterLab 3

vaibhavsagar:


jupyterlab-ihaskell/src/codemirror-ihaskell.ts
```diff
@@ -1,31 +1,35 @@
-import * as CodeMirror from 'codemirror';
+import { ICodeMirror } from '@jupyterlab/codemirror';
 
 import 'codemirror/lib/codemirror';
```
bollwyvl, line 3: you probably want to clobber _all_ these `lib/codemirror` imports...

jupyterlab-ihaskell/src/codemirror-ihaskell.ts
```diff
@@ -1,31 +1,35 @@
...
 import 'codemirror/mode/haskell/haskell';
+import 'codemirror/mode/r/r';
```
bollwyvl, line 5: likely would have to `Mode.ensure('r')` as well

vaibhavsagar: Updated, thanks!

jupyterlab-ihaskell/src/codemirror-ihaskell.ts
vaibhavsagar: Updated!

jupyterlab-ihaskell/src/codemirror-ihaskell.ts
```diff
@@ -1,31 +1,36 @@
...
-    },
-    {
-      open: /\[r\||\[rprint\||\[rgraph\|/,
-      close: /\|\]/ ,
-      mode: CodeMirror.getMode(config, "text/x-rsrc"),
-      delimStyle: "delimit"}
-  );
-});
+export function defineIHaskellMode({ CodeMirror }: ICodeMirror) {
+  Mode.ensure("haskell");
```
bollwyvl, line 7: hm, these are promises... might be a little racy, but you're _probably_ fine...

jupyterlab-ihaskell/src/index.ts
```diff
@@ -2,27 +2,22 @@ import {
...
   autoStart: true,
-  requires: [],
-  activate: (app: JupyterFrontEnd) =>
+  requires: [ICodeMirror],
+  activate: (app: JupyterFrontEnd, codeMirror: ICodeMirror) =>
   {
-    app.serviceManager.ready
-      .then(() => {defineIHaskell()});
+    defineIHaskellMode(codeMirror);
+    console.log('ihaskell codemirror activated');
```
bollwyvl, line 19: hmmm... does it really need the log? i mean, if you've documented, "it works if you look in the console and see..." then keep it, but otherwise...

jupyterlab-ihaskell/src/codemirror-ihaskell.ts
vaibhavsagar: How would you suggest doing it instead?

jupyterlab-ihaskell/src/index.ts
vaibhavsagar: Hmm, you're right that it's largely unnecessary. I can remove this.

vaibhavsagar: Fixed in https://github.com/gibiansky/IHaskell/commit/2638434cae3b5e9e12192a6ffdc60deedd0eb385.

jupyterlab-ihaskell/src/codemirror-ihaskell.ts
bollwyvl: `export async function defineIHaskellMode` and then `await Mode.ensure` in both places. Then the index.ts could `defineIHaskellMode(codeMirror).catch(console.warn)` (without making  _the_ calling function async).

vaibhavsagar: Fixed in https://github.com/gibiansky/IHaskell/commit/92027da0f3486525d07fd462e19bbdd44dc9218f, thanks!

PR #1273, opened 2021-09-11
Remove `stack --docker` development support

vaibhavsagar:


vaibhavsagar:
I'm not aware of anyone who actually uses this and it's not tested by CI so I think it's a good idea to get rid of this functionality altogether.

Issue #1210, opened 2021-01-27
`:set -package ...` not supported yet


teto:
I would like to load  some custom libraries in ihaskell. I've tried to load them with `:set -package-db ` which returns a `Warning: -package not supported yet` stemming from
https://github.com/gibiansky/IHaskell/blob/5b7e03b2caad17a51cb7490d66bf808e0e8b9d4a/src/IHaskell/Eval/Util.hs#L219

Any hope it can be supported or a workaorund ? Our libraries are built by bazel if that's of any help (so not easily convertible to nix packages) and we are using ihaskell through tweag/jupyterWith.

A previously similar issue https://github.com/gibiansky/IHaskell/issues/692 . Apparently you use the GHC api directly, hie-bios does pass all the flags so that should be possible.

vaibhavsagar:
It looks like the relevant section of the GHCi source is here: https://github.com/ghc/ghc/blob/11bdf3cdd6efb406839a0ebe33455908a66df805/ghc/GHCi/UI.hs#L3094-L3108. If you'd like to adapt it to IHaskell I would be happy to review and merge.

teto:
I've looked at ghc implementation for this and this looks non trivial as few symbols are exported and it would be hard to maintain this for different ghc implementations.
I would like to allow ihaskell to find the existing libraries instead (which is non-trivial because I have a mix of ghc and bazel-built libraries). 
So my question is how does ihaskell find the libraries ? Does it rely on `Ihaskell  --ghclib=<path>` exclusively ? or can I configure a ghc with different package databases ? The latter would be easier for me since my libraries are scattered across different folders.

vaibhavsagar:
It primarily uses `--ghclib` but I recently added support for package environment files in https://github.com/gibiansky/IHaskell/pull/1280, so you can use multiple package databases that way if you need. I chose not to add that support for GHC < 9.0 since nobody asked for it and if you want to do the same thing with `:set -package` that would be fine by me.

teto:
is that the file referenced by GHC_ENVIRONMENT ? that's super cool. I will run some tests with ghc8 in https://github.com/teto/IHaskell/pull/new/ghc8-env (will open PR if that's conclusive).

vaibhavsagar:
Yes, but I meant it for use with `cabal`, e.g. https://github.com/gibiansky/IHaskell/issues/1012#issuecomment-932889297.

PR #1281, opened 2021-10-06
build: add flake support

teto:
I wanted like to test and possibly address https://github.com/gibiansky/IHaskell/issues/1210.
nix flakes make many things easier, including the override of dependencies from the command line, without the need to edit nix files, check their hash etc so a whole lot faster (for instance when updating Ihaskell in jupyterWith).

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,171 @@
+{
+  "nodes": {
+    "flake-compat": {
+      "flake": false,
+      "locked": {
+        "lastModified": 1627913399,
+        "narHash": "sha256-hY8g6H2KFL8ownSiFeMOjwPC8P0ueXpCVEbxgda3pko=",
+        "owner": "edolstra",
+        "repo": "flake-compat",
+        "rev": "12c64ca55c1014cdc1b16ed5a804aa8576601ff2",
+        "type": "github"
+      },
+      "original": {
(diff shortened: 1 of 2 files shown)


vaibhavsagar:
Thanks for this PR! I don't know if I want to add support for flakes to IHaskell at this time.

vaibhavsagar:
Just realised: why is `hls` included? It doesn't look like it's used in `flake.nix`, is it safe to remove?

teto:
ha I wanted to extend the flake to provide a development environment with haskell-language-server (hls) since it's so useful. Do you use it ? I can make it so that running `nix develop` gets you into a development shell with hls.

NB: if you dont want to install nixUnstable, I can provide a compatibility layer that uses the flake as source of truth but that you can still run with a simple `nix-shell`: this is what https://github.com/edolstra/flake-compat/ is about.

vaibhavsagar:
We do support GHC 8.10.7, I just updated the `stack.yaml` and it tests correctly. What issue are you experiencing?

teto:
yep looks good now thanks, I've opened https://github.com/gibiansky/IHaskell/pull/1287 as a follow up. `nix build` buils ihaskell. 22 tests fails because of nix (like some folders are readonly etc) so I've just disabled them.

PR #1284, opened 2021-10-09
ihaskell: add `use-hlint` flag

vaibhavsagar:


vaibhavsagar:
I'm primarily adding this to simplify the process of updating IHaskell to work with a newer GHC, since it's easier than figuring out how to bump HLint each time.

Issue #1289, opened 2021-11-01
Plugins don't work


abarbu:
Sadly, plugins don't work. But in a confusing way!

There's a notebook showing the problem here: https://gist.github.com/abarbu/664c35e5140033a45af588434c70d694

And an explanation below.

Start with a notebook that has the `ghc-typelits-knownnat` package available.

```
:set -XFlexibleContexts -XFlexibleInstances
:set -XPolyKinds -XRankNTypes -XScopedTypeVariables -XTypeApplications -XTypeFamilies
:set -XTypeFamilyDependencies -XTypeInType -XTypeOperators -XUndecidableInstances
import Data.Proxy
import GHC.TypeLits
let f :: forall n . (KnownNat n, KnownNat (n+2)) => Proxy n -> Integer; f _ = natVal (Proxy :: Proxy n) + natVal (Proxy :: Proxy (n+2))
```

Will work. Then add:

```
let f :: forall n . KnownNat n => Proxy n -> Integer; f _ = natVal (Proxy :: Proxy n) + natVal (Proxy :: Proxy (n+2))
```

This will fail with 
```
Could not deduce (KnownNat (n + 2)) arising from a use of ‘natVal’
```

Now enable the plugin with

```
:set -fplugin GHC.TypeLits.KnownNat.Solver
```

And again run

```
let f :: forall n . KnownNat n => Proxy n -> Integer; f _ = natVal (Proxy :: Proxy n) + natVal (Proxy :: Proxy (n+2))
```

In GHCi this works because of the plugin. But in iHaskell it doesn't and errors the same way as before the plugin.

The confusing part is that adding a plugin that doesn't exist with

```
:set -fplugin GHC.TypeLits.KnownNat.SolverX
```

produces errors about plugins that don't exist. So at some level plugins are being loaded. But they aren't being used.

abarbu:
Looks like plugins are working after all with :set
In case anyone else finds this, plugin errors don't appear immediately like they do in GHCi.

abarbu:
Sadly, plugins don't work. But in a confusing way!

Start with a notebook that has the `ghc-typelits-knownnat` package available.

```
:set -XFlexibleContexts -XFlexibleInstances
:set -XPolyKinds -XRankNTypes -XScopedTypeVariables -XTypeApplications -XTypeFamilies
:set -XTypeFamilyDependencies -XTypeInType -XTypeOperators -XUndecidableInstances
import Data.Proxy
import GHC.TypeLits
[...]

PR #1291, opened 2021-11-02
Handle plugin initialization using :set -fplugin module

abarbu:
This addresses #1289

Previously plugins were (almost) silently ignored. I also got rid of a misleading comment.

Plugins work as expected now.

### DIFF
--- a/src/IHaskell/Eval/Util.hs
+++ b/src/IHaskell/Eval/Util.hs
@@ -49,6 +49,7 @@ import           GHC.Types.TyThing
 import qualified GHC.Driver.Session as DynFlags
 import qualified GHC.Utils.Outputable as O
 import qualified GHC.Utils.Ppr as Pretty
+import           GHC.Runtime.Loader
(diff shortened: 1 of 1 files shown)


vaibhavsagar (approved):
Approved contingent on tests passing!

vaibhavsagar:
IIRC you still have commit privileges so you could merge this yourself but I'm hoping you will fix the compilation failures on GHC <8.10 first :smile:. I would be fine with a solution that compiles but doesn't provide plugin loading functionality on older GHCs we support, I know getting everything working across multiple versions can be a huge pain.

vaibhavsagar:
This looks great, can you fix the test failure by changing the expected output to the new output?

abarbu:
Thanks! I incorporated those changes. Looks like everything should pass now.
I don't think I have commit access anymore, if you don't mind merging.

vaibhavsagar:
Thanks so much!

PR #1294, opened 2021-11-05
Replace `cereal` with `binary`

vaibhavsagar:


vaibhavsagar:
After reading [this Reddit comment from `glguy`](https://www.reddit.com/r/haskell/comments/qmpfa4/cerealinstances/hjbpl56/?utm_source=reddit&utm_medium=web2x&context=3) I thought I would look into how difficult it would be to make this change.

PR #1301, opened 2021-11-10
Remove imports

teto:
I was trying to hack on ihaskell but the build logs were littered with unused imports so I cleaned up a bit.

I fixed the issue with static-canvas. I can drop that commit if you prefer but considering there has been no release since 2019, I think this is an ok fix.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -13,3 +13,8 @@ packages: .
           ./ihaskell-display/ihaskell-plot
           ./ihaskell-display/ihaskell-static-canvas
           ./ihaskell-display/ihaskell-widgets
+
+source-repository-package
+  type: git
+  location: https://github.com/jeffreyrosenbluth/static-canvas.git
+  tag: 1aad0f192828ded72dced14e42ac4e6baa6dab6f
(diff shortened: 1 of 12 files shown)


vaibhavsagar:
Those imports were in there for a reason :smile:. If you want to remove them in a way that maintains backwards compatibility then I am happy to merge.

vaibhavsagar:
No, the CI failed because the imports you removed were required for GHC 8.2.

vaibhavsagar:
I agree that the current `fail-fast` behaviour is unhelpful though, so I've made https://github.com/gibiansky/IHaskell/pull/1302 to change it.

teto:
ah I see the stack 8.2 failure triggered all other jobs to be cancelled ? is there a way to disable that from the .github actions (couldnt find it). Even if I restore some of the imports there are probably others that are safe to remove.

Also what's the policy to drop ghc compatibility ?

vaibhavsagar:
There's no policy for GHC compatibility, but I would prefer to keep support for older GHCs for now.

vaibhavsagar:
Just merged #1302, feel free to rebase for more helpful CI output.

vaibhavsagar (approved):

vaibhavsagar:
Please rebase to get rid of the `fixup!` commit.

teto:
I've updated the flake so that it can propose different ihaskell versions (instead of just ghc8107).
I've just rebased.

vaibhavsagar:
Thanks!

PR #1306, opened 2021-11-28
improve flake by distinguishing different goals (development vs final executable vs with a jupyter environment)

teto:
When working with the flake I realised a few issues:
- I need some special tools just in dev
- some commands won't work without jupyter so let's bring it in

I took inspiration for the naming from 
https://github.com/haskell/haskell-language-server/blob/b05e14d7fbd0d600ec8e687b0ea29ad38a39341c/flake.nix#L237

I am still unsure about haskell-env vs jupyter-env or anything else. Better name it right straightaway since changing this afterwards would be a pain for 3rd party projects like jupyterWith.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -87,11 +87,11 @@
         "pre-commit-hooks": "pre-commit-hooks"
       },
       "locked": {
-        "lastModified": 1633827357,
-        "narHash": "sha256-TjTkSYMjce1ejPVvidnFWluxij6LGZV5TSmx2IQPYX4=",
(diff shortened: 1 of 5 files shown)


vaibhavsagar (approved):
LGTM!

vaibhavsagar:
I don't really have an opinion on how to name the development environment vs. executable, but the rest of the changes look good.

teto:
let it be ihaskell-env then feel free to merge ^^

PR #1313, opened 2022-01-06
feat(nix): add darwin support in flake

teto:
not enabling it in ihaskell make nix flake check fail for jupyterWith.

Not sure if that will work though.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -14,7 +14,7 @@
   };
 
   outputs = { self, hls, nixpkgs, flake-utils, ... }:
-    flake-utils.lib.eachSystem ["x86_64-linux"] (system: let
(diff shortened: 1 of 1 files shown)


teto:
ie. , to fix the failure visible at https://github.com/tweag/jupyterWith/runs/4732925096?check_suite_focus=true 

the ihaskell doesn't test for mac builds here. I dont own any so I have no interest in it but if someone from jupyter knows how to add it easily that could be interesting for the project.

vaibhavsagar:
You're right that it would be a good idea to have macOS builds as well, it's a bunch of work that I'm not particularly excited about doing right now but noted that you want this.

Issue #1316, opened 2022-01-16
Instructions on developing IHaskell using Nix


FPtje:
## Request

I'm on NixOS. How do I get started with contributing to this project? i.e. How do I get the following to work:

- Get the haskell-language-server to work
- Build the project using stack and/or cabal

The `flake.nix` seems to have some infrastructure for a development environment, but it doesn't quite get me there.

### Background

The [README](https://github.com/gibiansky/IHaskell#linux) has some instructions on compiling the software on Linux. Since I'm using NixOS, and I don't have many of the dependencies installed, I figured I would use the `flake.nix` to start working on it. After all, the `flake.nix` contains a dev shell that seems to have most of the required dependencies.

However, I'm running into some issues. Here are my steps:

```bash
$ nix develop
$ stack
stack: command not found
# makes sense, I don't have it installed globally, and it's not in flake.nix.
$ cabal build
Warning: The package list for 'hackage.haskell.org' is 934 days old.
Run 'cabal update' to get the latest list of available packages.
Resolving dependencies...
cabal: Could not resolve dependencies:
[__0] trying: ghc-parser-0.2.3.0 (user goal)
[__1] trying: ghc-8.10.7/installed-8.10.7 (dependency of ghc-parser)
[__2] trying: base-4.14.3.0/installed-4.14.3.0 (dependency of ghc-parser)
[__3] trying: ihaskell-aeson-0.3.0.1 (user goal)
[__4] trying: here-1.2.13 (dependency of ihaskell-aeson)
[__5] next goal: haskell-src-meta (dependency of here)
[__5] rejecting: haskell-src-meta-0.8.2, haskell-src-meta-0.8.1,
haskell-src-meta-0.8.0.3 (conflict: ghc =>
template-haskell==2.16.0.0/installed-2.16.0.0, haskell-src-meta =>
template-haskell>=2.8 && <2.15)
[__5] rejecting: haskell-src-meta-0.8.0.2 (conflict: ghc =>
base==4.14.3.0/installed-4.14.3.0, haskell-src-meta => base>=4.6 && <4.12)
[__5] rejecting: haskell-src-meta-0.8.0.1, haskell-src-meta-0.8 (conflict: ghc
=> base==4.14.3.0/installed-4.14.3.0, haskell-src-meta => base>=4.6 && <4.11)
[__5] rejecting: haskell-src-meta-0.7.0.1, haskell-src-meta-0.7.0 (conflict:
[...]

PR #1317, opened 2022-01-16
Add jupyter to flake.nix dev shell

FPtje:
## Summary

Part of fixing https://github.com/gibiansky/IHaskell/issues/1316
Adds `ipython`, `jupyter` and `jupyter-lab` to the environment of the dev shell. This way, there's definitely a jupyter notebook to run the IHaskell kernel with.

My laptop doesn't have this globally installed, so I think this is something the dev env might want to provide. Especially because it is not possible to run `ihaskell install` without it.

When people do have it installed globally, I don't reckon this will conflict too much. Though I'm not sure what happens if the version in the `flake.nix` is newer than the globally installed one. If it makes backwards incompatible changes to the files in a user's data directory that would be unfortunate. I'm not sure if Jupyter does that :thinking: 

It's an open question, feel free to reject this PR if it is considered inappropriate to provide those dependencies in here :+1:

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -36,6 +36,8 @@
       ghc8107  = pkgs.haskell.packages.ghc8107;
       ghc921  = pkgs.haskell.packages.ghc921;
 
+      pythonDevEnv = pkgs.python3.withPackages (p: [p.jupyterlab]);
+
       mkDevShell = hsPkgs:
         let
           myIHaskell = (mkPackage hsPkgs);
@@ -44,6 +46,7 @@
           myModifier = drv:
             pkgs.haskell.lib.addBuildTools drv (with hsPkgs; [
               cabal-install
+              pythonDevEnv
               self.inputs.hls.packages.${system}."haskell-language-server-${compilerVersion}"
             ]);
         in (myModifier myIHaskell).envFunc {withHoogle=true;};


vaibhavsagar:
This is a good point. `flake.nix` was contributed by @teto so I'd like to get his opinion before making a decision.

flake.nix
```diff
@@ -36,6 +36,14 @@
       ghc8107  = pkgs.haskell.packages.ghc8107;
       ghc921  = pkgs.haskell.packages.ghc921;
 
+      pythonDevEnv = pkgs.python3.withPackages (p: [
+        p.ipykernel
```
vaibhavsagar, line 40: All of these dependencies except `jupyterlab` are unnecessary: https://github.com/gibiansky/IHaskell/blob/6362b9b1d68e347749792c783b6bc12a33600d07/release.nix#L41

flake.nix
```diff
@@ -44,6 +52,7 @@
           myModifier = drv:
             pkgs.haskell.lib.addBuildTools drv (with hsPkgs; [
               cabal-install
+              pythonDevEnv
```
teto, line 55: in my local flake, I just added pkgs.jupyter to test the installation of kernels. If you want advanced testing facilities, you can run the flake at https://github.com/tweag/jupyterWith/ and use `--override-input ihaskell ~/myihaskell` for instance.

vaibhavsagar: Hmm, it sounds like it would be good to include this by default. Would that cause any issues for you?

teto: I was planning to submit a PR with that, but I dont know ifit's enough for OP's purpose.

vaibhavsagar (changes requested):
Please update `pythonDevEnv` to only include `jupyterlab`.

flake.nix
FPtje: I don't have much of a preference here. It sounds like a good idea :+1:

FPtje:
The `pythonDevEnv` has been updated to only include `jupyterlab` :+1:

vaibhavsagar (approved):
LGTM!

vaibhavsagar:
Thank you!

jamesdbrock:
Thanks for all this @FPtje

PR #1321, opened 2022-01-30
ghc 9.2.1 support in flake

teto:
WIP, opening this to prevent duplicate work.


vaibhavsagar:
I think you will need to update https://github.com/gibiansky/IHaskell/blob/c78143d4a4e3669ddf55724498ca39a9c8585b88/.github/workflows/nix.yml#L37-L39 to reflect these changes as well.

.github/workflows/nix.yml
```diff
@@ -34,9 +34,9 @@ jobs:
     strategy:
       matrix:
         versions:
-          - release: 'release-8.10.nix'
-          - release: 'release-9.0.nix'
-          - release: 'release-9.2.nix'
+          - release: 'release-8107.nix'
+          - release: 'release-90.nix'
```
vaibhavsagar, line 38:
```suggestion
          - release: 'release-902.nix'
```

vaibhavsagar:
Test failures look weird, the `ghc902` attribute is missing and there is a hash mismatch in `nixpkgs` for the 9.2.1 build, but if you fix those everything else looks good to me!

teto:
Would it be ok to go for flake-only support ? I am a big fan of flakes that solve 80% of the issues with nix. And I am not sure the pre-flake release.nix are really used/practical anyway. I am mentiong that since nix build .#ihaskell-XXX seem to work locally but their 

```
        nix-build --keep-going \
        -I nixpkgs=https://github.com/NixOS/nixpkgs/tarball/nixos-21.11 \
        ${{ matrix.versions.release }} \
```
counterpart doesnt.

I find it impractical to test each ghc set with nixpkgs=https://github.com/NixOS/nixpkgs/tarball/nixos-21.11, I would rather have each ghc version use an appropriate nixpkgs version.

Maybe we could have the  "main" ghc version track the latest stable nixos version like https://github.com/NixOS/nixpkgs/tarball/nixos-21.11 but for instance ghc921 should be able to track a more recent nixpkgs. What do you think ?

vaibhavsagar:
I use `release.nix` :smile:. When we introduced `flake.nix` in https://github.com/IHaskell/IHaskell/pull/1281 it was not discussed that this would involve getting rid of `release.nix`. I don't think `flake.nix` covers all the use cases of `release.nix` yet (e.g. the IHaskell display packages), so I think we should keep it for now.

I agree that using `nixos-21.11` for each GHC is not practical, I'm open to changing CI to build each ghc version with an appropriate nixpkgs.

> Maybe we could have the "main" ghc version track the latest stable nixos version like https://github.com/NixOS/nixpkgs/tarball/nixos-21.11 but for instance ghc921 should be able to track a more recent nixpkgs. What do you think ?

Sounds good to me.

teto:
> I use release.nix 

haha good ! Just got frustrated because the release.nix contain redundant code. I will need to think about how to rearchitect the files to reduce the redundancy between the flake and the release*.nix . After a good night of sleep though :)

flake.nix
```diff
@@ -61,29 +63,76 @@
...
+          packageOverrides = pfinal: pprev: {
+            openapi-core = pprev.openapi-core.overridePythonAttrs(oa: {
+
+              doCheck = false;
+              doInstallCheck = false;
+            });
+          };
+        };
+
+        ihaskellGhc921 = prev.haskell.packages.ghc921.extend (final: prev: let
```
vaibhavsagar, line 107: This looks like it duplicates a bunch of the functionality in `release.nix`.

vaibhavsagar, line 107: Also there's a reason we use `.override` in `release.nix` instead of `.extend` even though the latter is more convenient: https://github.com/NixOS/nixpkgs/issues/26561#issuecomment-354862897

release-902.nix
```diff
@@ -0,0 +1,27 @@
+let
+  nixpkgs-src = builtins.fetchTarball {
+    url = "https://github.com/NixOS/nixpkgs/tarball/30d3d79b7d3607d56546dd2a6b49e156ba0ec634";
+    sha256 = "0x5j9q1vi00c6kavnjlrwl3yy1xs60c34pkygm49dld2sgws7n0a";
+  };
+  overlays = [ (final: prev: {
+    python3 = prev.python3.override {
+      # Careful, we're using a different self and super here!
+      packageOverrides = pfinal: pprev: {
```
vaibhavsagar, line 9: Could we resolve this issue by using a newer nixpkgs?

flake.nix
teto: good point

teto:
Thanks for the review. A few points:
1/ any idea what can I do to fix the CI stack errors ? looks unrelated
2/ after trying out this version in my project, I realized it would bring 2 ghc and it didn't compose well. So I've ended up doing what I usually do in other projects, aka expose the nix code in an overlay so that other projects can create their ghcenv with ihaskell in it.
3/ I am a ihaskell beginner and it's not clear to me what `ihaskell kernel` does. I've tried  
```
ihaskell kernel /home/teto/.local/share/jupyter/kernels/haskell/kernel.json 
ihaskell: ihaskell: /home/teto/.local/share/jupyter/kernels/haskell/kernel.json: Failed to parse profile file
```
with /home/teto/.local/share/jupyter/kernels/haskell/kernel.json being the one created by `ihaskell install`
```
{"argv":["/nix/store/mv8q4sa8lh851hab97d3nqvqxmzc5mbb-ihaskell-0.10.2.2/bin/ihaskell","kernel","{connection_file}","--debug","--conf","/home/teto/.ihaskell/rc.hs","--ghclib","/nix/store/51qcvp5gv869dwgd5jn4wg81zf1ynjw5-ghc-9.2.1/lib/ghc-9.2.1","+RTS","-M3g","-N2","-RTS"],"display_name":"Haskell","language":"haskell"}
```
How can I fix that issue ? I am trying to redure my test loop (I need to import haskell modules built by bazel so it's quite hackish at the moment). Currently I start jupyter-lab and run my commands from there so it's quite slow (I have to restart stuff to test). If I could run `ihaskell run 'import Proprietary.Module'` to check my hacks work, that would help.

vaibhavsagar:
1) You're right, it's unrelated. I think it's the same as https://github.com/d2l-ai/d2l-book/issues/46, am investigating now and hopefully will have a fix soon. If I'm unable to fix it before you're done with this PR we can ignore those failures and merge anyway.

2) That sounds fine, if you want to introduce a `/nix` directory and split things out there I'm fine with that too. I'm not attached to the version-specific `release-*.nix` files, but I would like to keep `release.nix` for now at least. If you want to split the logic in `release.nix` out into a bunch of different files I'm willing to consider that too.

3) `ihaskell kernel` is the subcommand that sets up a connection between IHaskell and Jupyter. It's used by Jupyter directly as part of the kernel spec: https://jupyter-client.readthedocs.io/en/stable/kernels.html#kernel-specs

thomasjm:
This can be closed I think now that #1425 is merged

teto:
I've wanted to update this but it was tough. thanks for your PR, I will try testing it soon.

PR #1343, opened 2022-05-10
Use gtk2hs-builtools-0.13.8.3

soppelmann:
IHaskell currently doesnt build on AArch64 machines due to gtk2hs-buildtools being outdated, see https://github.com/gtk2hs/gtk2hs/issues/308

Using the latest version solves this problem.


vaibhavsagar:
Thanks!

Issue #1358, opened 2022-07-18
avoid switching directory


SmartHypercube:
Is it possible to avoid all silent directory-switching? My code is operating files by relative paths, and I find it annoying that every time I draw a chart the working directory is changed to /tmp by this:
https://github.com/IHaskell/IHaskell/blob/5bcfb8976cbd4dbdf0ee6e38b2c0d0021068b354/ihaskell-display/ihaskell-charts/IHaskell/Display/Charts.hs#L30

I don't think silently switching directory in any library code could be a good idea. The best approach may be to completely avoid using a temporary file, i.e. directly render to `ByteString`. If that is not possible, you could use a temporary file without switching to /tmp. Just use the absolute path to access it. If switching directories is impossible to avoid, please use [`withCurrentDirectory`](https://www.stackage.org/haddock/lts-19.15/directory-1.3.6.2/System-Directory.html#v:withCurrentDirectory). It has a shortcoming though: code running in other concurrent threads will be affected.

SmartHypercube:
If a temporary file is needed, please use [`withTempFile`](https://www.stackage.org/haddock/lts-19.15/temporary-1.3/System-IO-Temp.html#v:withTempFile) because:
- it allocates a unique file name for each use, so that two threads/processes using temporary files concurrently will not conflict.
- it doesn't just find an unused file name and tell you to use it (which leads to race conditions), instead it creates and opens the file to keep other concurrent threads/processes from using it.
- it deletes the file after use.

daviddavo:
The [`chart-cairo`](https://hackage.haskell.org/package/Chart-cairo-1.8/docs/Graphics-Rendering-Chart-Backend-Cairo.html) package doesn't export any functions that return a bytestream.

This library uses the functions [`withPDFSurface`](https://hackage.haskell.org/package/cairo-0.13.3.0/docs/Graphics-Rendering-Cairo.html#t:withPDFSurface), `withSVGSurface`, `surfaceWriteToPNG`, etc. which write the output to a file... I tried looking at the source, but it goes very low level.

The easiest thing should be using a tmp file, but it's true that this function presents race conditions, as the filename is always the same, and it changes the directory for every thread in this process. That should be certainly changed.

SmartHypercube:
I too failed to find a way to render to bytes with chart-cairo before when I used it in a project.

[...]

PR #1359, opened 2022-07-19
avoid switching directory in ihaskell-display

SmartHypercube:
related: #1358 

I removed the function `switchToTmpDir` and replaced every usage of it with `withSystemTempFile`.


vaibhavsagar:
Thanks so much!

PR #1360, opened 2022-07-19
Dynamic way for dynamically linked RTS Haskell distros

YakoYakoYokuYoku:
Fixes issues with the failure of discovery of platform libraries in Haskell distributions where the RTS was linked dynamically (like Arch Linux for example).

### DIFF
--- a/src/IHaskell/Eval/Util.hs
+++ b/src/IHaskell/Eval/Util.hs
@@ -9,6 +9,7 @@ module IHaskell.Eval.Util (
     setExtension,
     ExtFlag(..),
     setFlags,
+    setWayDynFlag,
 
     -- * Code Evaluation
     evalImport,
@@ -42,6 +43,7 @@ import           GHC.Driver.Monad (modifySession)
 import           GHC.Driver.Ppr
 import           GHC.Driver.Session
 import           GHC.Driver.Env.Types
+import           GHC.Platform.Ways
 import           GHC.Runtime.Context
 import           GHC.Types.Name (pprInfixName)
 import           GHC.Types.Name.Set
@@ -58,6 +60,7 @@ import           GHC.Driver.CmdLine
 import           GHC.Driver.Monad (modifySession)
 import           GHC.Driver.Session
 import           GHC.Driver.Types
+import           GHC.Driver.Ways
 import           GHC.Types.Name (pprInfixName)
 import           GHC.Types.Name.Set
 import qualified GHC.Driver.Session as DynFlags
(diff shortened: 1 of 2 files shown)


vaibhavsagar:
Thanks for this patch! I'm happy to merge when tests pass, but also there isn't anything in our CI to check that this works so it might break again in the future. Do you have any suggestions on how we might add such a check?

vaibhavsagar:
Hmm, it also looks like GHC 9.2 doesn't have a `GHC.Driver.Ways` module and it has been replaced by [`GHC.Platform.Ways`](https://downloads.haskell.org/~ghc/latest/docs/html/libraries/ghc-9.2.2/GHC-Platform-Ways.html).

YakoYakoYokuYoku:
> Hmm, it also looks like GHC 9.2 doesn't have an `addWay'`.

Happens that I've overlooked the detail that there's an `addWay` function, though I haven't seen it since the API docs at Hackage for GHC 9.2 are not currently available.

vaibhavsagar:
They are here: https://downloads.haskell.org/~ghc/latest/docs/html/libraries/ghc-9.2.2/index.html

YakoYakoYokuYoku:
> Thanks for this patch! I'm happy to merge when tests pass, but also there isn't anything in our CI to check that this works so it might break again in the future. Do you have any suggestions on how we might add such a check?

Does this repo has a Haskell CI environment that uses dynamic only packages, i.e. every package is built with `--enable-shared` only, and not dynamic-too neither static only? If so, then that image can be used for checks. Otherwise another image can be created for it.

vaibhavsagar:
No, we do not have such an environment already, otherwise we would have discovered this issue earlier :smile:.

vaibhavsagar:
Hmm, `nix-build release-9.0.nix --arg enableStaticLibraries false` works for me on current `master`, so I don't think this is right.

vaibhavsagar:
I'm merging this for now, with the caveat that we currently have no way to check this works and will continue to work in the presence of changes.

YakoYakoYokuYoku:
> I'm merging this for now, with the caveat that we currently have no way to check this works and will continue to work in the presence of changes.

Well, in Arch Linux this is required because every Haskell package is built with `-dynamic`, so the `setWayDynFlag` function (by the usage of `addWay'`) _appends_ `-dynamic` to a GHC session. The `hie-bios` package also does this with [`setWayDynamicIfHostIsDynamic`](https://hackage.haskell.org/package/hie-bios-0.9.1/docs/HIE-Bios-Ghc-Gap.html#v:setWayDynamicIfHostIsDynamic) so it can be used in those kind of distributions too. I can't wrap my mind around with how to test it in Nix, but AFAICT this works for Arch Linux, maybe that could be used unless there's the way to doing it in Nix :thinking:.

Issue #1367, opened 2022-09-06
New ghc-parser release?


thomasjm:
Currently `ghc-parser-0.2.3.0` on Hackage has an upper bound of `ghc < 9.1`. This prevents IHaskell from being built on the latest `nixpkgs-unstable` branch against GHC 9.2.

It looks like the current in-tree `ghc-parser` has been updated to support 9.2. Could you tag a new `ghc-parser` release and put it on Hackage? Thanks!

vaibhavsagar:
I've uploaded new versions of `ghc-parser`, `ipython-kernel`, and `ihaskell`.

thomasjm:
Awesome, thanks!

PR #1368, opened 2022-09-11
Check in stack-9.2.yaml for GHC 9.2 build

thomasjm:
The builds fine for me, thought it might be worth checking into the repo to confirm GHC 9.2 support.

Also confirms #1367 should be easy to fix :)


vaibhavsagar:
Thanks!

PR #1369, opened 2022-09-12
A couple more stack files tweaks + update CI

thomasjm:
Follow up to #1368.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,6 +44,7 @@ jobs:
           - stack-yaml: 'stack-8.8.yaml'
           - stack-yaml: 'stack.yaml'
           - stack-yaml: 'stack-9.0.yaml'
+          - stack-yaml: 'stack-9.2.yaml'
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


vaibhavsagar:
Thanks again!

thomasjm:
No problem! Could I get you to look at #1367 @vaibhavsagar? It should be easy to do, and would unblock IHaskell on Nixpkgs for GHC 9.2.

vaibhavsagar:
Hey, I've been putting it off because I have to bump the version numbers and upload everything correctly, a tedious yet detail-oriented task. I will get to it in the next day or two, please feel free to ping me again if that doesn't happen

thomasjm:
Sounds good, thanks for looking into it!

Not sure I totally understand the difficulty -- `ghc-parser` is a standalone package and it doesn't seem like its version number is synced up with anything else (like `ihaskell` proper or any of the display packages). Isn't the whole task just to bump the version number to `0.2.3.1` (I don't think a major bump is even needed) and upload to Hackage? No pressure of course, just excited to get this working :)

vaibhavsagar:
It's not just `ghc-parser`, `ipython-kernel` and `ihaskell` will also need new uploads. My workflow for this could perhaps use improvement but since I do it so infrequently I don't really see the point in making it faster.

PR #1383, opened 2022-11-13
Add options to set kernel and display name

### LINKED ISSUE #710: Run more than one version of IHaskell?
Is it possible to build and run ihaskell on different GHC versions?  eg 7.10.1 and 7.10.3?  If you try doing so, each overwrites the other's `ihaskell` binary in `.local/bin`.  It's easy to save a copy, but the location seems to be hardwired in?

FPtje:
Fixes #710 

This adds the options `--kernel-name` and `--display-name` to the `ihaskell install` command. They respectively set the directory name and the display name of the installed IHaskell kernel

### Motivation

IHaskell always installs the kernel at the same location, with the same display name. With these options, we are free to create different IHaskell kernels. This is very useful if you have multiple projects, with different kernels each.

### Usage:

```
$ ihaskell install --help
Up to date
install [OPTIONS]
  Install the Jupyter kernelspec.

  -l --ghclib=<path>          Library directory for GHC.
     --use-rtsopts="<flags>"  Runtime options (multithreading etc.). See `ghc +RTS -?`.
     --debug                  Print debugging output from the kernel.
     --kernel-name=<name>     The directory name of the kernel.
     --display-name=<name>    The display name of the kernel.
  -c --conf=<rc.hs>           File with commands to execute at start; replaces ~/.ihaskell/rc.hs.
     --prefix=<install-dir>   Installation prefix for kernelspec (see Jupyter's --prefix option)
  -? --help                   Display help message
     --stack                  Inherit environment from `stack` when it is installed
     --env-file=<file>        Load environment from this file when kernel is installed

$ ihaskell install --kernel-name=haskell-myproject --display-name="Haskell (MyProject)"
```

Tested by running manually, starting a jupyter notebook and verifying that the installed kernel is selectable and runs Haskell code.


vaibhavsagar:
Thanks so much!!

PR #1384, opened 2022-11-14
Added instruction for apple silicon Macs 

bsong0:


vaibhavsagar:
These instructions specify using `cabal`, but we don't do that anywhere else. Can you update the instructions to use `stack` for consistency?

vaibhavsagar:
Hmm, it's confusing to have the `cabal` instructions not use the cloned repo, unlike the `stack` instructions. I'm going to close this for now, since the updates are related to upstream issues with XCode and Stack as opposed to issues in IHaskell we need to fix.

PR #1392, opened 2022-12-31
Add -dynamic back in; removal broke IHaskell

abarbu:
This commit removed `-dynamic` as a default flag https://github.com/IHaskell/IHaskell/commit/2891bc8498ce12ae2677d72c7252e2623158e378

The results of this change break IHaskell in two ways. First, it now links every package available statically on startup. This is fine for small projects. In a large project though, with 200 packages, I gave up after 15 minutes. Second, many of us use IHaskell with packages that rely on shared libraries. This stopped working entirely.

There's a related long discussion here
https://github.com/IHaskell/IHaskell/pull/1252/files#diff-30f4411ae6041ebe629c055ae5c4407e47989f719a5733416bfa23454270ba65R130

The suggested solution is that users should add dynamic in this case. But there is simply no way for users to figure this out.

I had to run strace and notice among its 100Mb of output that ihaskell is stuck loading one object file after another and look at gdb to confirm that it's stuck in the linker pinning the CPU to 100%. No one is going to do this and then realize that -dynamic is needed in IHaskell.

But it's even worse for a small project. You will inexplicably get errors about shared libraries. And they'll be lazy! So most of the code works, until you trip over some code that uses the shared library and then you get an incredibly cryptic error about symbols missing. Users have absolutely no way of tracking this back to IHaskell, never mind realizing that they need to add -dynamic.

This was done to make building docker images simpler. https://github.com/IHaskell/IHaskell/pull/1252

All of this is a horrible cost to pay for some convenience when building docker images. It breaks the common case for a slightly more convenient corner case. The downside of this pull request is it breaks building the Docker images, would you mind fixing that @brandon-leapyear? You seemed to have a version that copies over the right files.

I suggest we add -dynamic back in. @junjihashimoto suggested the same in the discussion above.

### DIFF
--- a/ihaskell.cabal
+++ b/ihaskell.cabal
@@ -136,7 +136,11 @@ executable ihaskell
   hs-source-dirs: main
   other-modules:
                    Paths_ihaskell
-  ghc-options: -threaded -rtsopts -Wall
+  -- NB Without -dynamic IHaskell loads every package statically on startup
+  -- This makes startup incredibly slow on large projects
+  -- Without -dynamic IHaskell also cannot load shared libraries
+  -- This leads to users getting cryptic error messages about missing symbols
+  ghc-options: -threaded -rtsopts -Wall -dynamic
 
   if os(darwin)
     ghc-options: -optP-Wno-nonportable-include-path


brandonchinn178:
Hi please reference this account in the future; brandon-leapyear is my work account.

I haven't touched this project since the PR, so I don't really have any motivation for fixing this. Just like any OSS project, anyone should make the changes they want; just cuz I rewrote the Docker stuff, doesnt mean I'm forever the maintainer of the code. Of course, feel free to completely revert the PR if you dont want to bother with getting dynamic working with the new Docker implementation (which I suggested in the original PR too; I was never forcing this rewrite in)

I never had a version that copied over the right files. To get it working with the new multi-stage build, you'd have to add a COPY for all `.so` files linked to by ihaskell, which would be annoying now, and whenever ihaskell gets a new transitive dependency. I really dont think docker is the best solution for a dynamic library; I might recommend hiding the -dynamic flag behind a cabal flag and disable it in the docker build.

Just for my own knowledge, I still dont understand why ihaskell needs -dynamic, can you elaborate? My intuition would be that you build the ihaskell binary, you run it and specify a haskell project at runtime, and it shells out to ghc to compile and run the project. Why does running Haskell code at runtime depend on how the driver is linked?

abarbu:
> Hi please reference this account in the future; brandon-leapyear is my work account.

Sure! The PR was done under that account that's why I referenced it.

> I haven't touched this project since the PR, so I don't really have any motivation for fixing this. Just like any OSS project, anyone should make the changes they want ...

Sounds good! I think you're taking my comments in a very negative light. Sorry. They were never meant to be negative. We just want to use IHaskell :)

> I might recommend hiding the -dynamic flag behind a cabal flag and disable it in the docker build.

Then IHaskell will be broken in this unfortunate way for everyone who uses docker. But it is a low-effort compromise.

> My intuition would be that you build the ihaskell binary, you run it and specify a haskell project at runtime, and it shells out to ghc to compile and run the project. Why does running Haskell code at runtime depend on how the driver is linked?

Well, not quite. It doesn't 'shell out' to GHC, which would mean that GHC (well, GHCi) runs in a different process (although this would be one way of implementing IHaskell). It starts a GHC session within the IHaskell process.

Code for the installed packages needs to come from somewhere. They aren't linked into the ihaskell binary (otherwise you would need to recompile IHaskell every time you added a package). Without `-dynamic` every package is loaded with the GHC static linker, with `-dynamic` every package is loaded with the system dynamic linker. The latter is lazy and fast. And without the latter you don't get to load shared libraries.

Perhaps you're asking a different question: why do the compile-time flags of IHaskell determine its runtime behavior, couldn't the two be decoupled? Likely, yes, they could be. But it wouldn't help anything.

At runtime we want IHaskell to use the dynamic linker, nothing would be gained from having an IHaskell compiled without `-dynamic` that runs a GHC session that behaves as if `-dynamic` was turned on. You would still need to include all of those shared libraries for anything to work. Even if IHaskell was implemented by having GHCi run in a different process it wouldn't help, then we'd want that GHCi to use the dynamic linker and those shared libraries.

brandonchinn178:
> I think you're taking my comments in a very negative light. Sorry. They were never meant to be negative.

No worries, I didn't take them as negative 🙂 My intent was simply to reply to your "do you mind fixing the Docker images" question with "I do mind because I was never intending to forever maintain the Dockerfile" with as little harsh-ness as possible, which I possibly failed in.

> Code for the installed packages needs to come from somewhere. They aren't linked into the ihaskell binary (otherwise you would need to recompile IHaskell every time you added a package).

When you say "package", what do you mean here? Do you mean a dependency of IHaskell or a dependency of the project you're using IHaskell to run?

> Even if IHaskell was implemented by having GHCi run in a different process it wouldn't help, then we'd want that GHCi to use the dynamic linker and those shared libraries.

Ok so let's pretend IHaskell is compiled statically linked, and implemented such that it receives Haskell code from you, then runs `callProcess "ghc" ["-dynamic", "-e", <code>]`. It sounds like you're saying this is still an issue because the shared libraries _from the compilation of IHaskell itself_ don't exist to be reused. But my argument is that the build of IHaskell itself _should_ be sandboxed from the runtime user environment. e.g. I should be able to build the IHaskell binary and send it to you, and you should be able to run it, with IHaskell building and linking third-party dependencies built in the usual project sandbox (e.g. `~/.stack` or `~/.cabal/store`). Whether IHaskell _itself_ has the dependencies statically linked or is loading deps dynamically, the _Haskell evaluator it's driving_ should be completely separate from that.

That last paragraph was me speaking generally; this all seems particularly ill-suited for the Docker use-case. If your project uses `postgresql-libpq`, how do you install the libpq libraries in the Docker image for IHaskell to compile/run your code against? One option could be that each of your (non-trivial) projects should have its own `Dockerfile` that extends the IHaskell one:
```hs
FROM ihaskell
RUN apt-get install postgresql-dev
```
That is, the IHaskell Docker image's sole responsibility is to provide the IHaskell binary (and an OS that can run the binary) and it's up to you to set up the environment for your particular project. And again, this should be the case whether or not IHaskell is dynamically linked

vaibhavsagar:
I'm happy to change this back, since among other things the start-up time without `-dynamic` is a lot slower and restarting the kernel takes forever. In fact we default to doing it the old way when using Nix. However this change is currently breaking the Nix build so I need to fix whatever is happening there  before this can be merged.

vaibhavsagar:
I can't merge this in its current state, so closing for now.

PR #1414, opened 2023-08-06
Switch from pager to proper HTML output for code highlighting

thomasjm:
Currently, running commands like `:info` creates some pretty weird HTML output. It contains a huge chunk of CSS (much of it unnecessary), then a `textarea` containing the actual content. Worst of all, it then contains Javascript like this:

```javascript
<script>CodeMirror.fromTextArea(document.getElementById('code'), {mode: 'haskell', readOnly: 'nocursor'});</script>
```

All we want to do is highlight some Haskell code here. Relying on CodeMirror and Javascript to do so is brittle and non-portable, not to mention insecure--most sensible notebook implementations won't just let you run arbitrary Javascript.

As a result, the current approach needs to use the pager to work properly, which creates a subpar user experience, as documented in https://github.com/IHaskell/IHaskell/issues/1394.

This PR takes a different approach. Using the `ghc-syntax-highlighter` library, we parse the Haskell code within IHaskell, and then we output simple HTML with classes indicating the colors to use.

The PR adds 2 new options to IHaskell: `--html-code-wrapper-class` and `--html-code-token-prefix`, for controlling the way the classes are determined. Out of the box the classes should show up properly in Jupyter notebook/lab.


vaibhavsagar:
I appreciate the effort you put into this PR, but there are a couple of issues that prevent me from merging this PR in without substantial changes:
1. `ghc-syntax-highlighter` only works with GHC 8.4 onwards, so we would either need to use a different library, drop support for GHC < 8.4 (which is something I would want to discuss separately instead of doing it as part of this PR), or do some ungodly conditional dependency thing and keep both approaches (which sounds like the worst of both worlds).
2. Possibly related to the above, the tests don't pass.
3. It looks like you're plumbing things through to the workhorse `evaluate` command and its related commands, which seems smelly to me. IMO that's a pretty invasive change only to support HTML output, and I wonder if we could do it another way that makes use of existing kernel state/options.

In general I would really like it if you would open an issue or otherwise discuss large PRs like this before doing all of this work, especially because it puts me in an awkward position to have to say no or request substantial changes. I realise I've been unresponsive lately but a large contribution like this also creates additional work for me going forward when I have to:
- modify this code in response to changing requirements
- keep it up-to-date with newer GHC versions
- maintain backwards compatibility

I do agree that the current HTML output is far from ideal and I hope we can figure out something that works for everyone here.

thomasjm:
> In general I would really like it if you would open an issue or otherwise discuss large PRs like this before doing all of this work, especially because it puts me in an awkward position to have to say no or request substantial changes.

It's okay, I'm not trying to pressure you to merge. I'm going to use my fork for my own purposes whether this change gets upstreamed or not. I'm also happy to work with you until you're happy with it.

I think I need to bring a flake change over from another branch to fix the Nix test.

I'm very open to suggestions on how to structure the extra arg with the kernel options and handle GHC < 8.4.

vaibhavsagar:
I'm still thinking about the other issue, but in the meantime I've fixed the test suite so it's running again.

thomasjm:
> What if we stop highlighting code cells for GHC < 8.4? I'd be okay with that.

That sounds good! I was thinking the same, we could just provide plaintext output for older GHCs. I'll take a stab at that when I get a moment.

vaibhavsagar:
If you look at the definition of `KernelState`: https://github.com/IHaskell/IHaskell/blob/7de07ec3a3b11153ed23b83ce20df2a96229c92c/src/IHaskell/Types.hs#L181-L193

It already has configuration values for things such as `useSvg` and `kernelDebug`. I think it would be relatively straightforward for you to put your HTML options there too. What do you think?

thomasjm:
I did consider putting the options in the `KernelState`, but my thought process was as follows:

We want these options to be configurable as flags, which means they should be parsed alongside the other flags in `parseKernelArgs`:

https://github.com/IHaskell/IHaskell/blob/7de07ec3a3b11153ed23b83ce20df2a96229c92c/main/Main.hs#L91

Thus, the options need to go into `KernelSpecOptions` to start with. At that point I had two options,

1. Add the options to `KernelState` as well and copy them there when it gets constructed, or
2. Just plumb the `KernelSpecOptions` through to the evaluation function.

On balance I preferred 2), because 1) means putting the options in two separate data structures. Also, `KernelState` is supposed to be "state stored in the kernel between executions" -- it contains things that are meant to be mutable, via commands like `:option no-svg`. It seemed to me that these options are static configuration, not state. And there's no reason such configuration shouldn't be made available, and it may even make future configuration easier to add. I did consider whether hiding the plumbing in a reader monad might be nicer.

So yeah, that's my long-winded rationale! I'll make the change if you prefer of course.

vaibhavsagar:
I think I see where you're coming from, but IMO option 2) is more consistent with the way other configurable parameters work, e.g. `--debug` (which I don't think we can change when a kernel is running). I agree that separating static and dynamic configuration is conceptually cleaner but since we already aren't doing that AFAICT I don't think that starting here is a good idea.

thomasjm:
Okay, I've done the `KernelState` refactor and the selective HTMLification for GHC >= 8.4, and rebased on master.

I stuck a couple quality-of-life commits on the end, one to fix some Cabal warnings and one to make it easier to work with this on NixOS. (I've found that leaving out the `nix.enable` field in `stack.yaml` works better because then Stack will autodetect if it's running on NixOS and enable the flag automatically if so.) I can remove these commits or move them to separate PRs if you like.

I haven't tested the Stack build for GHC 8.0 and 8.2 yet due to errors like the one below. But let's see what the CI says...

```
ghc: panic! (the 'impossible' happened)
  (GHC version 8.0.2 for x86_64-unknown-linux):
        Prelude.chr: bad argument: 2634022915
```

vaibhavsagar:
Thanks for making those changes! I'm broadly happy with everything except the modifications to the test code, which seem a little strange to me (renaming etc. is fine, commenting tests out is not). I'll merge once we get the CI passing. One note on this: we'll probably want to disable the integration test for GHC 8.2 (it's already disabled for 8.0).

vaibhavsagar:
It looks like some of the test failures are due to `-Werror`, you're welcome to remove that from `stack.yaml` wherever it's an issue. It was introduced to improve code quality, but I personally think it's incompatible with how we support multiple GHCs, since imports e.g. differ so significantly across versions.

thomasjm:
Right, I still need to get those commented tests passing -- it was behaving flakily in a way that seemed to affect other tests.

I've been chipping away at the `-Werror` issues one CI run at a time. It's pretty tricky to test locally due to https://gitlab.haskell.org/ghc/ghc/-/issues/19452. I think it'll all be passing soon but I agree it's fragile. Maybe we could just disable the unused imports warning in CI but keep `Werror` for everything else. 

I was kind of thinking that if the behavior for different GHC versions is going to continue to diverge, it might make sense to split the integration test from being one big notebook file into many little notebooks, each with ~1 cell. Then it would be more straightforward to enable/disable specific ones for specific compilers, or even perhaps provide different expected outputs for different compilers. And it would make failures more granular and easy to diagnose.

vaibhavsagar:
It sounds like you might want to do some janky `git-worktree` thing in order to have everything in its own directory to avoid this issue, alternatively you could create `release.nix` files for each version you want to test. 

> I was kind of thinking that if the behavior for different GHC versions is going to continue to diverge, it might make sense to split the integration test from being one big notebook file into many little notebooks, each with ~1 cell. 

Yes, that seems like a good idea to me, although I don't see why the notebooks have to be 1 cell instead of many cells each like it is now. A related change I've been thinking about is completely getting rid of CPP in the codebase and using cabal conditionals à la http://blockscope.com/posts/2022-02-23-iffy-diffy.html, but it would be a huge undertaking and it wouldn't be as easy to fix bugs in the resulting codebase (since we'd have to do it in each version instead of in one place), so I'm undecided on whether it's a good idea.

thomasjm:
> It sounds like you might want to do some janky git-worktree thing in order to have everything in its own directory to avoid this issue, alternatively you could create release.nix files for each version you want to test.

Pretty sure that would only work if I also abstract the home directory, because I think some problematic interface files go into `~/.stack`. I do like the idea of doing more `release.nix` files though. I have a separate branch with some changes where I'm working on organizing all the `stack.yaml` files and `release.nix` into their own directories and making it all more systematic, if you're open to it.

> A related change I've been thinking about is completely getting rid of CPP in the codebase and using cabal conditionals

I haven't seen this approach before, looks promising. I think the first step in such an undertaking would be to refactor large modules like `IHaskell.Eval.Evaluate` into smaller pieces so as to isolate the different conditional paths. Maybe after doing that step the CPP wouldn't feel so unmanageable anymore and it wouldn't be necessary to actually go further...

src/tests/IHaskell/Test/Eval.hs
```diff
@@ -186,10 +200,10 @@ testEval =
         trace "test" 5
       |] `becomes` ["test\n5"]
 
-    it "immediately applies language extensions" $ do
-      [hereLit|
-        {-# LANGUAGE RankNTypes #-}
+    -- it "immediately applies language extensions" $ do
```
vaibhavsagar, line 203: Still need to re-enable this test.

vaibhavsagar:
> I have a separate branch with some changes where I'm working on organizing all the `stack.yaml` files and `release.nix` into their own directories and making it all more systematic, if you're open to it.

Definitely open to considering other approaches (in a separate PR please :smile:), although I think the root of the issue is that we don't have an explicit backwards compatibility guarantee and so I've been stubborn about supporting GHC 8.0 onwards. If we bit the bullet and e.g. dropped pre-9.0 support we would have a lot fewer files to worry about and it might not be as important to reorganise them. Again, that is something I would like to address separately to this.

It looks like you're pretty close to having this completed, I'm excited to merge it soon!

thomasjm:
I wonder if `DisplayData` could be given a less opaque `Show` instance? I'm guessing it doesn't have one now because the output can be arbitrarily long, but it makes it hard to debug the failures that are happening now. In a few cases these same tests are passing locally for me in the Docker container I jury-rigged to test GHC 8.x versions. If not a `Show` instance that prints everything, maybe one that truncates after a few hundred characters?

vaibhavsagar:
When I was trying to fix the tests earlier, I used the `ToJSON` instance to meaningfully represent `DisplayData` in `IHaskell.Test.Eval`. If you like you can include this in your changes:

```diff
diff --git a/ihaskell.cabal b/ihaskell.cabal
index dc840bb..4e08b81 100644
--- a/ihaskell.cabal
+++ b/ihaskell.cabal
@@ -179,6 +179,7 @@ Test-Suite hspec
         IHaskell.Test.Hoogle
     default-language: Haskell2010
     build-depends:
+        aeson,
         base,
         ghc,
         ghc-paths,
diff --git a/src/tests/IHaskell/Test/Eval.hs b/src/tests/IHaskell/Test/Eval.hs
index 8ec1310..638a651 100644
--- a/src/tests/IHaskell/Test/Eval.hs
+++ b/src/tests/IHaskell/Test/Eval.hs
@@ -4,6 +4,7 @@ module IHaskell.Test.Eval (testEval) where
 
 import           Prelude
 
+import           Data.Aeson (encode)
 import           Data.List (stripPrefix)
 import           Control.Monad (when, forM_)
 import           Data.IORef (newIORef, modifyIORef, readIORef)
@@ -47,7 +48,7 @@ becomes string expected = evaluationComparing comparison string
     comparison (results, _pageOut) = do
       when (length results /= length expected) $
         expectationFailure $ "Expected result to have " ++ show (length expected)
-                                                           ++ " results. Got " ++ show results
+                                                           ++ " results. Got " ++ show (encode results)
 
       forM_ (zip results expected) $ \(ManyDisplay [Display result], expect) -> case extractPlain result of
         ""  -> expectationFailure $ "No plain-text output in " ++ show result ++ "\nExpected: " ++ expect
```

src/tests/IHaskell/Test/Eval.hs
```diff
@@ -173,11 +157,37 @@ testEval =
...
+                        , DisplayData MimeHtml "<div class=\"code CodeMirror cm-s-jupyter cm-s-ipython\"><span class=\"cm-keyword\">type</span><span class=\"cm-space\"> </span><span class=\"cm-variable-2\">String</span><span class=\"cm-space\"> </span><span class=\"cm-atom\">::</span><span class=\"cm-space\"> </span><span class=\"cm-atom\">*</span><span class=\"cm-space\"><br /></span>\n<span class=\"cm-keyword\">type</span><span class=\"cm-space\"> </span><span class=\"cm-variable-2\">String</span><span class=\"cm-space\"> </span><span class=\"cm-atom\">=</span><span class=\"cm-space\"> </span><span class=\"cm-atom\">[</span><span class=\"cm-variable-2\">Char</span><span class=\"cm-atom\">]</span><span class=\"cm-space\"><br />  \t</span><span class=\"cm-comment\">-- Defined in \8216GHC.Base\8217</span><span class=\"cm-space\"><br /></span></div>"
+                        ]]
+        ]
+#elif MIN_VERSION_ghc(8,4,0)
+      displayDatasBecome ":in String" [
+        ManyDisplay [Display [
+                        DisplayData PlainText "type String = [Char] \t-- Defined in `GHC.Base'"
```
vaibhavsagar, line 175: It seems like most of the remaining test failures are because it's some gross unicode instead of the quotes around `GHC.Base`.

thomasjm: Yeah, I suspect it depends on locale and that's why it passes in my local Docker container.

vaibhavsagar: I don't know what the correct solution is here. We could explicitly specify the locale as part of the workflow, but then the tests would fail locally, which isn't what we want. Maybe we could try using [`setLocaleEncoding`](https://hackage.haskell.org/package/base-4.18.0.0/docs/GHC-IO-Encoding.html#v:setLocaleEncoding)? IMHO that's pretty janky also.

thomasjm: Hmm, I think doing `setLocaleEncoding` to put the locale in a known state within the tests is an okay thing to do.

It leaves the possibility that users with different locales will get different results. But if any of those results are bad somehow then another test suite could be added which tests different locales.

vaibhavsagar: I agree, I think for our purposes it will work fine.

src/tests/Hspec.hs
```diff
@@ -10,7 +11,9 @@ import           IHaskell.Test.Eval (testEval)
 import           IHaskell.Test.Hoogle (testHoogle)
 
 main :: IO ()
-main =
+main = do
+  setLocaleEncoding utf8
```
vaibhavsagar, line 15: Might be good to put a comment here explaining why we're doing this for when we inevitably forget a few months from now.

thomasjm: Will do if it works!

thomasjm: Hmm, this isn't producing identical behavior with my Docker container yet.

## CI

```
  src/tests/IHaskell/Test/Eval.hs:49: 
  1) Code Evaluation evaluates :in directive
       Expected display datas to be "[[{\"text/plain\":\"type String = [Char] \\t-- Defined in \195\162\194\128\194\152GHC.Base\195\162\194\128\194\153\"}]]". Got "[[{\"text/plain\":\"type String = [Char] \\t-- Defined in \226\128\152GHC.Base\226\128\153\"}]]"
```

## Docker

```
ihaskell              >   src/tests/IHaskell/Test/Eval.hs:49: 
ihaskell              >   2) Code Evaluation evaluates :in directive
ihaskell              >        Expected display datas to be "[[{\"text/plain\":\"type String = [Char] \\t-- Defined in \195\162\194\128\194\152GHC.Base\195\162\194\128\194\153\"}]]". Got "[[{\"text/plain\":\"type String = [Char] \\t-- Defined in `GHC.Base'\"}]]"
```

It hurts my brain a little to think about where this discrepancy could be arising...

thomasjm: Current hypotheses:

* GHC is called as a library within IHaskell somehow, maybe this doesn't respect the current process's locale and needs to be configured some other way?
* Something about how the logs are processed in GitHub Actions is changing it on the CI side?? 

Maybe next step is to try `LC_ALL=C` vs `LC_ALL=C.UTF-8`

thomasjm: I reverted that commit since it seems to have no effect.

src/tests/IHaskell/Test/Eval.hs
thomasjm: Unfortunately `setLocaleEncoding` doesn't seem to work (see my experiments below).

It seems the path of least resistance is just to just assume a UTF-8 locale, and expect UTF quotes in the tests. If a dev environment has a different locale for some reason, you can set `LC_ALL=C.UTF-8`.

With that change the eval tests are passing and we're almost there!

thomasjm: It's green :tada:

vaibhavsagar:
Merged, thanks so much for working on this!

thomasjm:
My pleasure, thanks for pushing through with me

jamesdbrock:
Thanks @thomasjm :heart:

PR #1425, opened 2023-11-28
Nix and stack.yaml organization

thomasjm:
Hi @vaibhavsagar -- I mentioned a while back that I'd been working on refactoring the Nix and stack.yaml files, and here is the result!

It looks like a lot of changes, but a lot of it is pretty mechanical. Here are the highlights:

* `flake.nix` and `*.nix` files have been completely refactored under the `nix/*` dir, and are much cleaner and more regular. For each supported GHC version, we define a Nixpkgs pin and a set of overlays which make it work. If you run `nix flake show`, you can see that we build `ihaskell` executables, plain environments, and display environments (with all the display packages included).  CI is currently passing for GHC 8.10 through 9.8.
* Stack files have been organized under the `stack/*` dir. All of them have their `nix` stanzas updated, so they work with `stack build --nix`. CI is currently passing for GHC 8.8 through 9.6. (EDIT: acceptance test failing on 8.8 only)
* Stack CI is much simpler now since it gets all its dependencies from Nix.
* Nix CI uses the same fixed Nixpkgs pins as the flake, for better reliability.

Caveats:

* ~~I've disabled the acceptance test part of the Stack CI, since it runs in the Nix one.~~
* GHC 9.8 is not fully ready:
  * The acceptance test is disabled since it doesn't yet support `hlint`, and
  * The display environment doesn't build, since some packages aren't 9.8-ready in the Nixpkgs master Hackage pin yet.


nix/overlay-9.6.nix
```diff
@@ -0,0 +1,35 @@
...
+            owner = "mrkkrp";
+            repo = "ghc-syntax-highlighter";
+            # 0.0.10.0
+            rev = "71ff751eaa6034d4aef254d6bc5a8be4f6595344";
+            sha256 = "wQmWSuvIJpg11zKl1qOSWpqxjp2DoJwa20vaS2KHypM=";
+          }) {};
+
+          ghc-lib-parser = self.ghc-lib-parser_9_6_3_20231014;
+
+          zeromq4-haskell = super.zeromq4-haskell.overrideAttrs (oldAttrs: {
```
vaibhavsagar, line 16: This was a workaround for a bug in a previous version of Cabal (https://github.com/haskell/cabal/issues/8455), but I think Nixpkgs has implemented its own workaround for now so we can probably get rid of this.

vaibhavsagar:
Wonderful, thanks for doing this! I'm broadly in favour of merging this, just need to more thoroughly review the changes. Also, do we need the `stack.yaml.lock` files or can we do without them?

vaibhavsagar:
That's fair, but we didn't have them before and IMO it makes an already large PR (in terms of files added/removed) even larger. I also think the supply chain attack scenario is somewhat unrealistic: most people using IHaskell + Stack will probably be using it through Stackage instead of through this repo directly. Can we move them to a separate PR?

nix/overlay-9.6.nix
thomasjm: Done

thomasjm:
Okay, I added a commit to remove them. We could squash this PR, or merge it as-is and then you can always just revert that commit to bring them back.

vaibhavsagar:
After giving it some thought, I don't think disabling the Stack integration test is a good idea. Changing the resolver frequently breaks the display packages because of dependency issues, and I think it's good to know when it breaks so we can implement workarounds in `stack.yaml` or in the codebase otherwise.

thomasjm:
:+1: Okay, I didn't have any particularly good reason for disabling it.

However, are you sure the display packages are the reason to do that? The display packages are still being built as part of the `stack build` command. Also, the acceptance test notebook doesn't test any display packages...

To me, the acceptance test is not very good -- it's brittle and not very granular, and often the diff it produces is hard to interpret to see why a test failed. I think its primary value is in simply confirming that the kernel works with real Jupyter stuff. My thought was that running it in the Nix pipeline is enough to confirm that. But I'd really like to get rid of it and replace it with fine-grained golden tests, using a CLI tool such as Papermill or perhaps `jupyter-console`.

vaibhavsagar:
Oh, you're totally right. I confused it with [the example notebook we used to use](https://github.com/IHaskell/IHaskell/blob/master/notebooks/IHaskell.ipynb). I still think it has value, and unless I'm misremembering I think it has helped us catch issues in the past as a smoke test. I'm very open to moving to Papermill etc., but I don't think `jupyter-console` tests everything we need it to, especially display-related aspects.

thomasjm:
Well here goes, I was having trouble getting the kernel to start before. Let's see if we can make it work...

.github/workflows/stack.yml
```diff
@@ -0,0 +1,103 @@
...
+    - uses: actions/cache@v3
+      name: Cache ~/.stack
+      with:
+        path: ~/.stack
+        key: ${{ runner.os }}-${{ matrix.stack-yaml }}-${{ hashFiles(format('./{0}', matrix.stack-yaml)) }}
+
+    - name: Build
+      run: |
+        export "NIX_PATH=nixpkgs=$(nix run .#print-nixpkgs-master)"
+        stack build --nix --stack-yaml ${{matrix.stack-yaml}}
```
vaibhavsagar, line 69: It's a bit janky but I think we want to copy the `stack-<version>.yaml` to the toplevel as `stack.yaml`, because it's what our Stack integration looks for during the integration test.

.github/workflows/stack.yml
```diff
@@ -0,0 +1,106 @@
...
+      if: ${{ !contains(fromJSON('["stack/stack-8.0.yaml", "stack/stack-8.2.yaml"]'), matrix.stack-yaml) }}
+      run: |
+        set -e
+
+        export "NIX_PATH=nixpkgs=$(nix run .#print-nixpkgs-master)"
+
+        nix build .#jupyterlab
+        export PATH="$(pwd)/result/bin:$(pwd)/.local/bin:$PATH"
+
+        stack install --nix --stack-yaml ${{matrix.stack-yaml}}
```
vaibhavsagar, line 91: Also why are we using the `--nix` option here?

.github/workflows/stack.yml
thomasjm: Done, though it doesn't seem to have helped with the current problem -- let's discuss below

thomasjm:
CC @teto who was working with IHaskell on Nixpkgs recently

vaibhavsagar:
As before, I broadly approve of the goals of this PR, but there are some aspects of it I'm a little uncomfortable with, and I think it's because this PR is really multiple PRs all rolled together:

- I'm not sure that refactoring all the Nix expressions to use flakes is a good idea for this project right now: I'm relatively ignorant of flakes myself, and this makes it substantially harder for me to make changes since I have no idea what I'm doing (and I can't guarantee that you or someone like you will always be around to make those changes for me). This seems like a separate change to something like moving the `*.nix` files under a `/nix` subdirectory without changing anything else. However, it is pretty nice that the Nixpkgs pins etc. all line up and are consistent, and since you preserved my `release.nix` file it shouldn't be too difficult for me to adapt in the short term.

- Using `stack --nix` is IMO the wrong approach: one of the goals of the current Stack CI is to test IHaskell the way a potential user would use it, which is why we have e.g. the otherwise unnecessary `requirements.txt` file. If we went all-in on using Stack with Nix, we would have to update our installation instructions to assume all our users are installing Nix too, which is a non-starter (although that would certainly make my life a lot easier). The fact that you had to extend IHaskell to pass through extra Stack flags to support this change also suggests we're getting far beyond file reorganisation.

I'm also a little confused about what you mean by "some packages aren't 9.8-ready in the Nixpkgs master Hackage pin yet": is this something I need to fix on my end, or can we disable the failing test for now until everything else catches up?

Here's what my ideal PR would look like:

1. Moving all the `*.nix` files into a `/nix` subdirectory (perhaps even leaving `release.nix` at the top level) and making the bare minimum changes required to get everything building and passing the Nix CI.
2. Moving all the `stack*.yaml` files into a `/stack` subdirectory (perhaps even leaving `stack.yaml` at the top level) and making the bare minimum changes to get everything building and passing the Stack CI.

I'm happy with the flakes refactoring and the code change to pass through Stack flags too, but not the `stack --nix` approach.

I'd love to discuss further over a voice or video call if that suits you better, I'm available at <first name><last name>@gmail.com.

vaibhavsagar (changes requested):
(see comment above)

thomasjm:
Sure, happy to discuss if we can find a time! Let me just briefly respond here:

> I'm also a little confused about what you mean by "some packages aren't 9.8-ready in the Nixpkgs master Hackage pin yet"

You can see that there are already a bunch of jailbreaks and version adjustments in the GHC 9.8 overlay just to build the core IHaskell package. I had started to add more to support the display packages, but ran into one (`diagrams-core`) that didn't even have the necessary version available in Nixpkgs master yet. So yes, I think we can just disable that test and wait for things to mature.

> I'm not sure that refactoring all the Nix expressions to use flakes is a good idea for this project right now

FWIW, I wouldn't say that this PR increases the use of flakes; there was already a flake before. It just went from messy and incomplete to clean and complete. To me, the flake aspect isn't too significant--it just provides a convenient way to manage Nixpkgs pins, and also to provide some useful Nixpkgs stuff that we can use in CI. As you say, it's fine to ignore the flake and just use the `release.nix` in the traditional way.

> Using stack --nix is IMO the wrong approach

Right, it seems like this is the main area we have to resolve. FWIW, I tried early on to keep the stack changes separate, but once I started Nixifying things it grew slightly entangled haha. I don't actually feel too strongly about it, but here's some thoughts on why I like it:

* The current CI builds things with Stack by installing a zoo of packages for Ubuntu 20.04. This may help users who use Ubuntu 20.04, but it doesn't mean much for users of other distros, or even other versions of Ubuntu. The unfortunate fact is, the display packages require a whole bunch of native dependencies, and installing those is always going to require some work for users outside of a few "blessed" platforms.
* Likewise, the `requirements.txt` may test things for the particular versions of `jupyter-*` installed in there, but that's a small comfort for users who have different versions. Anyway, Jupyter components are supposed to conform to a spec; that's how we ensure they work together. At most what this `requirements.txt` gives us is a smoke test, but using the `jupyter` from Nixpkgs works equally well for that.

Fundamentally, I just thought it was nice to say "all the IHaskell stuff builds cleanly against the libraries in the latest Nixpkgs release." Does it have to be this way? Of course not. We could equally well go back to Ubuntu and install the packages with `apt` instead. (I'd still suggest getting `jupyter` from Nixpkgs though; it seems unnecessary to do `pip` stuff in CI just for that.)

vaibhavsagar:
> FWIW, I wouldn't say that this PR increases the use of flakes; there was already a flake before

That's a great point :smile:, I'm sold on the flakes refactoring in this PR.

Everything you say about using Nix instead of `apt-get` etc. for system dependencies is true (it is more consistent and convenient) but I worry that it would negatively impact the user experience. If we went with this approach, the `requirements.txt` would become obsolete and we wouldn't be able to support it anymore or provide instructions for a hypothetical user to install IHaskell without first having Nix installed on their system. And if a user has Nix installed already, why bother with Stack support at all? I personally have no love for Stack, unfortunately my experience has been that most people who open issues with IHaskell use Stack, which is why I still want to support it. People also don't seem to take kindly to "just install Nix and try that approach" whenever I've suggested that as a resolution to their issue :sweat_smile:.

thomasjm:
> the `requirements.txt` would become obsolete and we wouldn't be able to support it anymore or provide instructions for a hypothetical user to install IHaskell without first having Nix installed on their system

Not at all, the instructions would just go from "install Jupyter using this `requirements.txt`" to "install Jupyter however you like, here's a link to Jupyter's own [instructions](https://jupyterlab.readthedocs.io/en/stable/getting_started/installation.html)." To me, Jupyter is a companion package that you need to use kernels like IHaskell--it's not incumbent upon IHaskell to teach users how to install Jupyter.

FWIW, even as an experienced of user of both IHaskell and Ubuntu, I wouldn't relish trying to get IHaskell + all the display packages working without using `stack --nix`.

Anyway, I just switched the CI back to use `apt`, and it's green! So we can go with that if you're okay with it.

(Just one idea: what if we had a separate CI pipeline that ran the `stack --nix` version? That way we could ensure the Nix stanza wouldn't bitrot for users who prefer to do it that way.)

BTW I tried to ping you on gchat a little while ago, not sure if you got the notification--happy to chat there if you like.

vaibhavsagar:
> (Just one idea: what if we had a separate CI pipeline that ran the stack --nix version? That way we could ensure the Nix stanza wouldn't bitrot for users who prefer to do it that way.)

Sure, that sounds great!

I'll go ahead and merge this now, the Docker build seems to be failing but I can fix that later.

> BTW I tried to ping you on gchat a little while ago, not sure if you got the notification--happy to chat there if you like.

Weirdly I didn't receive anything, I've sent you an email.

PR #1428, opened 2023-12-07
GHC 9.8 Nix display packages support

thomasjm:
The latest Nixpkgs `master` has what we need. Just need a bunch of overlays.


vaibhavsagar:
Thanks!

PR #1430, opened 2023-12-15
Use NixOS 23.11

vaibhavsagar:


vaibhavsagar:
@thomasjm does dropping 8.10 from the Nix build matrix cause any problems for you? I tried to edit the overlay but the `ghc-lib-parser`+`ghc-syntax-highlighter`+`hlint` version wrangling got too excessive for me. I also don't really see why we need to keep 9.0 and 9.2 there either, removing it would make CI faster and reduce the number of build artifacts we are caching which seems like a good idea to me since I don't know how many people are actually using those versions.

thomasjm:
It seems to me that 8.10 is a relatively alive compiler version still. 9.0 and 9.2 even more so.

Why not just add a separate pin? We can use 23.05 for older compilers and 23.11 for newer ones (and master for bleeding edge). The ability to do this was a major reason I organized the `mkVersion` function this way.

(Are we running into GitHub Actions limits with the caching or something?)

vaibhavsagar:
> (Are we running into GitHub Actions limits with the caching or something?)

I use Cachix for the Nix caching, and I've gotten multiple emails recently saying that I'm at my storage limit and garbage collection has started. Domen has kindly provided me with 20GB across both my personal cache (which uses <3.5GB with minimal traffic) and IHaskell's cache (which uses the remaining 16.5+GB and accounts for the bulk of bandwidth usage). I've always thought of it as a way to eliminate unnecessary builds for CI + provide a cached build of the latest IHaskell, and using it to provide prebuilt display environments for older GHCs falls outside that scope.

> It seems to me that 8.10 is a relatively alive compiler version still. 9.0 and 9.2 even more so.

What does "alive" mean here? 8.10 was [first released in 2020](https://www.haskell.org/ghc/blog/20200324-ghc-8.10.1-released.html), over 3 years ago and closer to 4. Even 9.2 was [first released over 2 years ago](https://www.haskell.org/ghc/blog/20211029-ghc-9.2.1-released.html). The current GHC bundled in `nixos-23.11` is 9.4, and it is likely that soon Stackage will switch to GHC 9.6.

> Why not just add a separate pin? We can use 23.05 for older compilers and 23.11 for newer ones (and master for bleeding edge). The ability to do this was a major reason I organized the mkVersion function this way.

I'm unclear about what the purpose of having Nix derivations for these older compilers is (and by "older" I mean that they are older than the `ghc` bundled with a particular NixOS revision). My policy (which I realise I've never stated explicitly, but until now there has been no need to do so) has always been that the Stack CI covers older supported GHC versions (since using the nightly resolvers is a pain in the butt and for the latest GHC versions it often doesn't exist at all) whereas there is always a Nix expression for the current `ghc` bundled in the latest revision of `nixpkgs` and all newer versions of GHC that are supported (because this is sometimes the only feasible way of building a package like IHaskell against a particular version of GHC).

I think the general question here is: what is IHaskell's CI for? When I started maintaining the project there were only Stack builds, and I decided to add Nix CI because I was beginning to use Nix professionally and personally. Soon it became obvious that supporting Nixpkgs `master` was a fool's errand, so I decided to stick to the latest `NixOS` revision as a set of somewhat stable packages in the vein of a Stackage LTS. At some point we added a Docker build so we could run IHaskell in a regularly-updated Docker container easily. After your PR we've arrived at the present day, where we are running multiple redundant (for the purposes of verifying that IHaskell builds against a particular version of GHC/Stackage LTS) builds per push/PR: Stack, Stack + Nix, Nix, Nix + display packages. This means a slower feedback loop on PRs, spurious failures (e.g. the current failing Stack+Nix GHC 8.8 check), increased cache pressure, and generally a waste of time and energy if nobody is consuming those build artifacts. It is for this reason that I would like to drop the pre-9.4 Nix checks: they're already well exercised by the Stack CI (twice!).

Also you didn't answer my question: are you using the 8.10 Nix binaries, or do you know someone who does that would justify the effort of having these checks run on *every single* PR?

thomasjm:
> What does "alive" mean here?

Just my general impression that many projects (and companies) are still happily using older (sometimes much older) versions of GHC. I don't have data to back this up though.

> Also you didn't answer my question: are you using the 8.10 Nix binaries, or do you know someone who does that would justify the effort of having these checks run on every single PR?

Oh sorry, I see. I don't use the IHaskell Cachix cache for anything and I had actually forgotten it exists. I don't care at all about prebuilt binaries. I don't really think anyone should[^1].

However, I definitely care about checking that IHaskell builds successfully with Nix against a healthy range of GHC versions.

I don't see the build time issue as too important--I didn't find it to slow down the feedback loop when I was working on my recent PR. I was using GitHub actions on my forked repo, and they ran fast enough for me. (Even without access to your Cachix key, I think.) Nor have I encountered any spurious failures--what failures there are (like with GHC 8.8) are un-debugged as of yet, but they seem to happen the same way every time. Maybe CI could be sped up by not testing the "plain IHaskell" environments, if we can keep the "IHaskell + display packages" ones green. Or, they both variants could happen in the same workflow job in different steps--that would give us the same speedup and also make it clear which one(s) failed.

In my ideal world, we'd
a) Keep 8.10 around using a 23.05 pin, and
b) Disable the `cachix/cachix-action` part of the CI workflow for old versions--disable them all for all I care :P

[^1]: After all, to run IHaskell, you already need a copy of GHC sitting around. Once you've downloaded that, compiling IHaskell from source doesn't take too long.

vaibhavsagar:
> Just my general impression that many projects (and companies) are still happily using older (sometimes much older) versions of GHC. I don't have data to back this up though.

This is a reasonable position, and in fact it is why we still have support stretching back to GHC 8.0. However, that comes with its own challenges, and frankly I think dropping support for 8.2 and older (at the very least) is long overdue. This is why I'm hesitant to open this particular can of worms, because it comes with the question of when it is appropriate to drop support for this particular combination of IHaskell and Nixpkgs. The maintenance burden of keeping these things around forever is not trivial, and I don't actually think it provides any value to anyone to know that e.g. 1 year from now we have IHaskell + GHC 8.10 + NixOS 23.05 working correctly, even though there's a mountain of overrides required to get it working with 24.11. That's also why I don't agree with

> Keep 8.10 around using a 23.05 pin

IMO if it can't easily build on the latest NixOS release, it doesn't need to be there at all.

> Oh sorry, I see. I don't use the IHaskell Cachix cache for anything and I had actually forgotten it exists. I don't care at all about prebuilt binaries.

I personally use the cache quite a bit, which might be why I'm harping on this one particular point.

> Disable the `cachix/cachix-action` part of the CI workflow for old versions--disable them all for all I care :P

Unfortunately it's not that simple, since doing that would make the checks take even longer, unless we do something else clever for caching (in which case, why not provide those to everyone?).

I think going forward I'm going to drop the Nix build for any GHC older than the default in the Nixpkgs we use, since that's what I was doing before.

thomasjm:
> something else clever for caching

This thing sounds perfect:

https://nix.dev/tutorials/nixos/continuous-integration-github-actions.html#caching-builds-using-github-actions-cache

https://github.com/DeterminateSystems/magic-nix-cache-action/

It makes me sad to lose 8.10 Nix support, especially since Nixpkgs 23.11 still provides GHC 8.10. (In fact, it provides versions going back to 8.6.) If we solve the caching another way and the overlays aren't too crazy, can we bring them back?

vaibhavsagar:
1. There needs to be a sane plan for when we drop support for older GHC versions.
2. I think we could probably bring them back now if you feel strongly about it, but as a separate repository that doesn't impact CI times on this repository.

vaibhavsagar:
> If we solve the caching another way and the overlays aren't too crazy, can we bring them back?

This highlights another interesting point, which is that Nixpkgs itself doesn't support older GHCs in the way you're proposing we do here, it merely provides them and leaves it to us to figure out the correct package versions. The fact that this is so difficult suggests to me that it's a bad idea and we shouldn't be doing it, but if you really want to I think we could probably figure something out.

thomasjm:
> There needs to be a sane plan for when we drop support for older GHC versions.

There must be data somewhere out there about how many people are currently using each version. From a purely CI and caching perspective, I think there's a lot of stuff we can do to improve the situation and keep supporting them if we want to.

> Nixpkgs itself doesn't support older GHCs in the way you're proposing we do here, it merely provides them and leaves it to us to figure out the correct package versions

Well, at any given time Nixpkgs has a "default" GHC version. The subset of Hackage that Nixpkgs has at that time is more or less designed to work with that GHC, so you need almost no overlays for it. As you go up or down in version number from the default, the expected overlay size gets bigger. That's why I think it's good to keep older pins around to support older GHCs: those older version of Nixpkgs are tuned for them. (BTW: now that this PR is landed, there is likely some unnecessary stuff in some of the overlays.)

An alternative that might be worth considering is [haskell.nix](https://github.com/input-output-hk/haskell.nix), which I've also used quite a bit. If we used that, we could do Nix builds that pull their dependencies straight from the `stack.yaml` files, so we don't have to manage overlays at all. I guess this has to be weighed as part of an overall strategy--for me, it might be better to stay with the Nixpkgs infra because it might be good to move some IHaskell builds into Nixpkgs itself.

vaibhavsagar:
> There must be data somewhere out there about how many people are currently using each version. From a purely CI and caching perspective, I think there's a lot of stuff we can do to improve the situation and keep supporting them if we want to.

Frankly I don't think enough people are using IHaskell for this to be anywhere near the amount of effort involved

What do you mean by "move some IHaskell builds into Nixpkgs itself"? Nixpkgs usually has a working IHaskell (e.g. [right now](https://github.com/NixOS/nixpkgs/blob/ee1148c372368108eead87063b2a43bf014e03dd/pkgs/development/haskell-modules/configuration-common.nix#L2769-L2772)) but it moves at a different pace than we do, so unless I'm mistaken about what you mean I don't think moving things into Nixpkgs will solve anything.

> An alternative that might be worth considering is [haskell.nix](https://github.com/input-output-hk/haskell.nix)

I don't want to use `haskell.nix`, it seems like overkill to me.

thomasjm:
> Frankly I don't think enough people are using IHaskell for this to be anywhere near the amount of effort involved

I'm happy to take a stab at the magic cache action thing, it looks easy to set up.

> What do you mean by "move some IHaskell builds into Nixpkgs itself"?

Nixpkgs currently has an IHaskell build for a single GHC. There's a bit of an initiative going on now to improve the Jupyter support in Nixpkgs and this could involve doing one for each of the supported GHCs. There's no reason Nixpkgs updates can't move at roughly the same pace as IHaskell itself.

vaibhavsagar:
I forgot to mention an issue with the magic cache approach: we're already using the GitHub Actions cache to (try to) cache the Stack builds, so it would have to be a separate repository. I'm not sure if there are per-organisational limits but I hope not.

PR #1429, opened 2023-12-07
CI: add stack --nix workflow

thomasjm:
Add `stack --nix` workflow, as discussed in #1425.

PR #1446, opened 2023-12-20
Bring back Nix 9.0, 9.2 using magic cache

thomasjm:
This demonstrates using the `DeterminateSystems/magic-nix-cache-action` in place of Cachix for certain builds, and brings back GHC 9.0 and 9.2 in the Nix CI to use it. 

It's fast, with a full Nix workflow build taking 2-3 minutes in my tests.

Hopefully this is a satisfactory way to maintain Nix support for these compilers without using too many Cachix resources.

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -33,18 +33,33 @@ on:
 jobs:
   build:
     if: (github.event_name != 'pull_request' && ! github.event.pull_request.head.repo.fork) || (github.event_name == 'pull_request' && github.event.pull_request.head.repo.fork)
+    name: "${{matrix.env.name}} (Cachix: ${{matrix.env.cachix}})"
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
         env:
-        - ihaskell-env-ghc94
-        - ihaskell-env-ghc96
-        - ihaskell-env-ghc98
+        - name: ihaskell-env-ghc90
+          cachix: false
+        - name: ihaskell-env-ghc92
+          cachix: false
+        - name: ihaskell-env-ghc94
+          cachix: true
+        - name: ihaskell-env-ghc96
+          cachix: true
+        - name: ihaskell-env-ghc98
+          cachix: true
 
-        - ihaskell-env-display-ghc94
-        - ihaskell-env-display-ghc96
-        - ihaskell-env-display-ghc98
+        - name: ihaskell-env-display-ghc90
+          cachix: false
+        - name: ihaskell-env-display-ghc92
+          cachix: false
+        - name: ihaskell-env-display-ghc94
+          cachix: true
+        - name: ihaskell-env-display-ghc96
+          cachix: true
+        - name: ihaskell-env-display-ghc98
+          cachix: true
 
     steps:
     - uses: actions/checkout@v3
@@ -53,19 +68,24 @@ jobs:
       with:
         install_url: https://releases.nixos.org/nix/nix-2.19.2/install
 
+    # Cache with either Cachix or the magic Nix cache, depending on the matrix.
+    # The latter uses only GitHub Actions storage.
     - uses: cachix/cachix-action@v12
+      if: ${{ matrix.env.cachix }}
       with:
         name: ihaskell
         signingKey: '${{ secrets.CACHIX_SIGNING_KEY }}'
+    - uses: DeterminateSystems/magic-nix-cache-action@main
+      if: ${{ !matrix.env.cachix }}
 
-    - name: Build environment ${{matrix.env}}
+    - name: "Build environment ${{matrix.env.name}} (Cachix: ${{matrix.env.cachix}})"
       run: |
-        nix build .#${{matrix.env}}
+        nix build .#${{matrix.env.name}}
 
-    - name: Check acceptance test for ${{matrix.env}}
+    - name: Check acceptance test for ${{matrix.env.name}}
       # Disable GHC 9.8 here since we don't have hlint support yet
       # Also, don't bother running it with the display envs since we already run it with the
       # basic envs, and it doesn't test any display stuff.
-      if: ${{ !contains(fromJSON('["ihaskell-env-ghc98"]'), matrix.env) && !contains(matrix.env, fromJSON('"display"')) }}
+      if: ${{ !contains(fromJSON('["ihaskell-env-ghc98"]'), matrix.env.name) && !contains(matrix.env.name, fromJSON('"display"')) }}
       run: |
-        nix build .#checks.x86_64-linux.${{matrix.env}} -L
+        nix build .#checks.x86_64-linux.${{matrix.env.name}} -L
--- a/flake.nix
+++ b/flake.nix
@@ -31,6 +31,8 @@
         };
         in
           pkgsMaster.lib.listToAttrs [
+            (mkVersion nixpkgs23_11  "ghc90"  [(import ./nix/overlay-9.0.nix)]  {})
+            (mkVersion nixpkgs23_11  "ghc92"  [(import ./nix/overlay-9.2.nix)]  {})
             (mkVersion nixpkgs23_11  "ghc94"  [(import ./nix/overlay-9.4.nix)]  {})
             (mkVersion nixpkgsMaster "ghc96"  [(import ./nix/overlay-9.6.nix)]  {})
             (mkVersion nixpkgsMaster "ghc98"  [(import ./nix/overlay-9.8.nix)]  { enableHlint = false; })
@@ -103,6 +105,6 @@
 
       defaultPackage = self.packages.${system}.ihaskell-env-ghc94;
 
-      devShell = self.packages.${system}.ihaskell-dev-ghc810;
+      devShell = self.packages.${system}.ihaskell-dev-ghc94;
     });
 }
--- a/nix/overlay-9.0.nix
+++ b/nix/overlay-9.0.nix
@@ -1,0 +1,14 @@
+sel: sup: {
+  haskell = sup.haskell // {
+    packages = sup.haskell.packages // {
+      ghc90 = sup.haskell.packages.ghc90.override {
+        overrides = self: super: {
+          singletons-base = self.callHackage "singletons-base" "3.0" {};
+          singletons-th = self.callHackage "singletons-th" "3.0" {};
+          th-desugar = self.callHackage "th-desugar" "1.12" {};
+          Chart-cairo = self.callHackage "Chart-cairo" "1.9.3" {};
+        };
+      };
+    };
+  };
+}
--- a/nix/overlay-9.2.nix
+++ b/nix/overlay-9.2.nix
@@ -1,0 +1,15 @@
+sel: sup: {
+  haskell = sup.haskell // {
+    packages = sup.haskell.packages // {
+      ghc92 = sup.haskell.packages.ghc92.override {
+        overrides = self: super: {
+          singletons-base = self.callHackage "singletons-base" "3.1" {};
+          singletons-th = self.callHackage "singletons-th" "3.1" {};
+          singletons = self.callHackage "singletons" "3.0.1" {};
+          th-desugar = self.callHackage "th-desugar" "1.13.1" {};
+          Chart-cairo = self.callHackage "Chart-cairo" "1.9.3" {};
+        };
+      };
+    };
+  };
+}


vaibhavsagar:
I just made a whole bunch of changes today to get the cache size down and reduce CI times, and this would undo a lot of that work. I'm still unconvinced that this is a good idea, did you explore doing it in a separate repository like I mentioned the last time we discussed this?

thomasjm:
Let me do a little experiment to clear the full cache on my fork and run all the workflows, and see how much cache it uses...

vaibhavsagar:
> But then what? Manually maintain two IHaskell repos going forward?

I was thinking something along the lines of having a workflow in a separate repository that clones this repo and runs the checks you want, which would obviate the need to maintain two separate repositories (the only thing we would have to maintain is that workflow).

It occurs to me that I haven't given you the full context here:

- Each repository on GitHub has 10GB of space available for caching. AFAICT there is no way to increase this number.
- I'm not willing to spend money on an alternative caching solution, because I don't want to be out of pocket for an open source project where I'm already contributing time and effort. I'm also not willing to implement my own caching on top of something like S3, which would in any case involve paying a cloud provider.
- After your recent PR we were routinely blowing past the cache limit, which means inconsistent and long GitHub Actions turnaround times since one evicted cache slows down the whole group of checks. This was due to a bad caching implementation on my part, since caching all of `~/.stack` is unnecessary and wasteful, but having extra caches for the `stack --nix` builds exacerbated the issue.
- I spent the vast majority of today on improving the Stack CI, largely through more granular caching (https://github.com/IHaskell/IHaskell/pull/1443), using different keys for `stack` and `stack --nix` builds (https://github.com/IHaskell/IHaskell/pull/1439) and avoiding redundant builds on `push`/`pull_request` events (https://github.com/IHaskell/IHaskell/pull/1440).
- After all this effort, we're sitting at around 8GB of cache utilisation and I would very much like to keep it comfortably under the 10GB limit.
- As far as I'm concerned anyone who wants to use a) IHaskell, b) Nix, and c) GHC < 9.4 can write their own expression to do so and there's no reason to add those expressions to this repo 

With all that in mind, I really really really really don't want to merge this PR. All day while I was working on this repo I was thinking how nice it would be if I disabled `stack --nix` CI as well and got that cumulative hour (spent waiting for it) of my life back.

thomasjm:
I totally understand the desire to save money and time on IHaskell.

> As far as I'm concerned anyone who wants to use a) IHaskell, b) Nix, and c) GHC < 9.4 can write their own expression to do so and there's no reason to add those expressions to this repo

I don't understand why you think GHC 9.0/9.2 (still popular and widely used) and Nix (growing in popularity) don't deserve to be supported. If I had to choose between the `stack --nix` CI and 9.0/9.2 on the Nix CI, I'd choose the latter. (If I knew it would mean sacrificing Nix support, I wouldn't have suggested running `stack --nix` at all.) Also, I'm kind of shocked by the suggestion that users would have to come up with these expressions--I'm kind of an expert at this at this point, and it takes me a while to figure it out for each GHC version.

BTW I'm trying to figure out exactly what contribution each of these workflows has on the cache--it's taking forever because clearing the cache entries seems to happen one entry at a time, but I should have results in a day or two.

I think the double repo thing could work. But there might be some easier answers too. What about making certain workflows (maybe the `stack --nix` and `Nix` ones) only run on manual dispatch? Then you could have quick feedback during day-to-day work by only running the Stack one. That's probably plenty to test for most regressions. Then, when it comes time to do a release, you can run the extra workflows and wait an hour.

vaibhavsagar:
> I don't understand why you think GHC 9.0/9.2 (still popular and widely used) and Nix (growing in popularity) don't deserve to be supported.

You don't have to sell me on Nix, I've been using it professionally since 2017 :smile:. What do you mean by "don't deserve to be supported"? We already test IHaskell against both these GHC versions on Stack and Stack+Nix. My issue is that I can't imagine the hypothetical user who wants this particular combination and does not have the ability to come up with it on their own. And even in that extremely unlikely case I would expect them to raise an issue on the repo.

Also, where does it end? IIRC you mentioned that 23.05 has GHC builds back to 8.6, so to be consistent we would want IHaskell expressions for those too? I've already fallen into the trap of supporting older GHC versions for far too long (you should see the issue someone raised to tell me off when I dropped GHC < 8) and I don't want to make the same mistake again.

> BTW I'm trying to figure out exactly what contribution each of these workflows has on the cache--it's taking forever because clearing the cache entries seems to happen one entry at a time, but I should have results in a day or two.

I was going to ask if you can access the https://github.com/IHaskell/IHaskell/actions/caches, but I just had a look at it and it's a mess because the magic nix cache seems to create a different entry for each derivation which makes it impossible to easily determine how much space is being used.

This isn't really about space usage though, although that is a component of it. It seems to me like you're seeing additional ways to build IHaskell as an unmitigated good, and it really isn't. As I have said before, every additional check incurs a cost on each commit and additional code to maintain over time, and it's unclear to me how many people are even e.g. using the Nix expressions we bundle at all. Note that there have been a grand total of 0 issues asking how to migrate from the old `release.nix` to the flake, which means one of 3 things:
1. Nobody is using IHaskell
2. Nobody is using the Nix expressions
3. Nobody needs help despite the sweeping change
In each of these cases, in my capacity as maintainer I don't think it's appropriate to pile on even more for some hypothetical good when weighed against the very real time I spend watching paint dry as the checks complete.

> I think the double repo thing could work. But there might be some easier answers too. What about making certain workflows (maybe the stack --nix and Nix ones) only run on manual dispatch? Then you could have quick feedback during day-to-day work by only running the Stack one. That's probably plenty to test for most regressions. Then, when it comes time to do a release, you can run the extra workflows and wait an hour.

I was thinking along the same lines too, but in the past we used to e.g. run the Docker build on a schedule and it was a huge pain to figure out what went wrong if/when that check failed. I might investigate something similar for the `stack --nix` builds but the Nix builds make good use of caching and are what I use to develop locally anyway and the Stack builds guard against regressions in older GHCs so I think it makes sense to keep both of those.

> IHaskell is normally a fairly stable project, and often sees months go by without a commit. Even if CI did take an hour to complete, is that so bad if it allows IHaskell to provide generous support for different GHC versions and building methods?

Harsh but fair. Have you considered that waiting for an hour for a one-line change would be deleterious to one's motivation to work on a project? I think I've been pretty good about adding support for new GHC versions and merging changes from other contributors, but otherwise I see my role as largely keeping things working smoothly. I've also recently found myself using IHaskell less and in the meantime JupyterLab has changed out from under me with https://github.com/IHaskell/IHaskell/issues/1417 and https://github.com/IHaskell/IHaskell/issues/1419, which means that I foresee myself (or someone else) needing to make substantial changes to get things working correctly again. Yet another reason to prioritise fast CI times.

Anyway, I don't mean to be dismissive, but this PR seems like a white elephant and I don't see it being a good idea to merge.

thomasjm:
I'll respond to this tomorrow after some sleep, but for now here's my first results. All tests are using the current state of this PR and an empty cache.

Building the **Nix CI workflow** takes **1h 2m 19s** and puts **436MB** in the cache.

Building the **Stack --nix CI workflow** takes **TODO** and puts **TODO** in the cache.

Building the **Stack CI workflow** takes **TODO** and puts **TODO** in the cache.

thomasjm:
> We already test IHaskell against both these GHC versions on Stack and Stack+Nix.

Stack+Nix is hardly a Nix build. It's only using Nix to obtain native dependencies. Stack is still running the show.

> the hypothetical user who wants this particular combination and does not have the ability to come up with it on their own

Like I said, coming up with it on your own is quite hard. Maybe the reason IHaskell doesn't seem to have many Nix users is because it's been so hard to use thus far?

> Also, where does it end? IIRC you mentioned that 23.05 has GHC builds back to 8.6, so to be consistent we would want IHaskell expressions for those too?

I definitely think a common-sense tradeoff can be made. At the current moment in time I think 8.6 or 8.8 might be sensible minimum versions. More support is a good thing for users if the balance of cache time/CI time/code complexity can be managed. I don't think anyone can complain about the project being practical.

> in the past we used to e.g. run the Docker build on a schedule and it was a huge pain to figure out what went wrong if/when that check failed

Well scheduled builds definitely sound like a bad idea, for the reasons you mention--and they may pollute the cache.

> Harsh but fair. Have you considered that waiting for an hour for a one-line change would be deleterious to one's motivation to work on a project?

It's not an insult! Stable projects are good. Of course we should try to make you as happy as possible with your workflow. FWIW, I rarely use CI for feedback because I can just run the tests on my local machine--for most changes, it's sufficient to test a couple GHC versions that I already have cached, and then check back on CI later on just to confirm it didn't break any random older versions. I find this to be faster than even the best CI.

---

Anyway: my experiments above are done now and they seem to show we only need `2.5GB` out of `10GB` to run all the CI workflows. Maybe your recent work on improving cache usage has been more successful than you thought?

vaibhavsagar:
> Stack+Nix is hardly a Nix build. It's only using Nix to obtain native dependencies. Stack is still running the show.

You're absolutely right, but my point is that we *do* test that 9.0 and 9.2 build correctly and pass all our tests. I would much rather wait for a third party to ask on the issue tracker for us to provide Nix expressions for these versions before I add them. This also raises another question: what is the value of the Stack+Nix checks if they're doing largely the same thing as the Stack builds?

> I definitely think a common-sense tradeoff can be made. At the current moment in time I think 8.6 or 8.8 might be sensible minimum versions.

If you look at other Haskell developer tooling projects, they're extremely conservative with which GHC versions they support: often only 2 or 3 major versions behind the latest one. A good example is [HLS](https://haskell-language-server.readthedocs.io/en/latest/support/ghc-version-support.html), which IIRC only supports 9.2 and above. `ghci`, which is more similar to IHaskell than HLS is, isn't backwards compatible with older GHC versions at all. In this respect, IHaskell is already an extreme outlier. Given that we already test all the versions that we do I don't want to expand the build matrix even further, especially with no reasonable policy around when to drop support. There's also no concrete evidence that having these expressions will make anyone's lives easier.

> FWIW, I rarely use CI for feedback because I can just run the tests on my local machine--for most changes, it's sufficient to test a couple GHC versions that I already have cached, and then check back on CI later on just to confirm it didn't break any random older versions. I find this to be faster than even the best CI.

I hope you don't think I just push code to a branch without running the tests first, but I would also like CI to be fast because especially when making changes to support a newer version of the GHC API I often mess up the CPP stanzas in a way that is difficult for me to test on my local machine.

It's been over 6 years now since I inherited this project, and in that time unfortunately I have accepted contributions that I have later regretted. This very strongly feels like one of those contributions: you contribute some code, I accept it, everything works and we're both happy right now; but in six months, a year, or two years from now I'm stuck maintaining it while you're off to bigger and better things. I'm not willing to put myself in that situation again.

When we first discussed this I said we could do it as a separate repository with a reasonable policy around when to drop support for older GHC builds. The current approach meets neither of those criteria.

thomasjm:
> what is the value of the Stack+Nix checks if they're doing largely the same thing as the Stack builds?

Their sole purpose is to keep the `nix` stanza of the `stack.yaml` from bitrotting. For that reason I think they don't at all need to run on every push. A very easy tweak would be to run them less frequently, perhaps only as part of a release process.

> There's also no concrete evidence that having these expressions will make anyone's lives easier.

They would make my life easier... especially with the project to integrate more IHaskell kernels into Nixpkgs.

> This very strongly feels like one of those contributions: you contribute some code, I accept it, everything works and we're both happy right now; but in six months, a year, or two years from now I'm stuck maintaining it while you're off to bigger and better things. I'm not willing to put myself in that situation again.

FWIW, I have no plans to stop being invested in IHaskell.

I find this kind of an unfair characterization of this PR:

* It is less than 30 lines of actual overlay code, which restores functionality to 2 whole GHC versions with Nix
* It shows how to combine Cachix with the magic Nix cache in order to lighten the load on Cachix, which is something I understood that you wanted.
* Per the numbers posted above, we're currently using 20% of the available cache before this PR. If accepted, this PR would take it to 25%. Still plenty of headroom if you ask me.
* If you combine this with my other two suggestions, you would have an even leaner and faster CI. To be clear, these are
  * Don't run the Stack+Nix pipelines on every push, and
  * Drop super old GHC versions from the Stack pipeline

vaibhavsagar:
> A very easy tweak would be to run them less frequently, perhaps only as part of a release process.

Hackage heavily discourages any sort of automated release process, which is why we don't have one (source: I am a Hackage trustee). This is a non-starter. The only option is to run them on a schedule, which has problems we've discussed above.

> They would make my life easier... especially with the project to integrate more IHaskell kernels into Nixpkgs.

I don't really see how: if that project takes off, the additional expressions are going to make it into Nixpkgs. We don't do anything fancy on our end when building `IHaskell` that would make including the expressions here a priority AFAIK. It's possible I'm missing something here.

> FWIW, I have no plans to stop being invested in IHaskell.

That's great to hear, but I can't rely on that. Things change, sometimes suddenly and unexpectedly.

> It is less than 30 lines of actual overlay code, which restores functionality to 2 whole GHC versions with Nix

This is disingenuous. As I have pointed out before, these are additional checks that run on every commit, and additional code that someone will have to maintain over time indefinitely.

> It shows how to combine Cachix with the magic Nix cache in order to lighten the load on Cachix, which is something I understood that you wanted.

I must have been unclear, I never wanted this. The increase in cache utilisation is one of multiple reasons I was against testing Nix expressions for older GHC versions, as I have been from the beginning.  The magic Nix cache is very cool tech but completely pointless if you want to put the built artifacts in a binary cache for other people to use. I think we have differing opinions on the utility of this, but my position is that if it is generally useful then it should be in a publicly available binary cache. I don't think these additional Nix expressions need to be in the cache, but that is because I don't think they need to be tested at all.

> Per the numbers posted above, we're currently using 20% of the available cache before this PR. If accepted, this PR would take it to 25%. Still plenty of headroom if you ask me.

That's only after I spent most of the day yesterday working on this. Not the most compelling argument if you ask me :smile:.

My numbers are substantially different:
```
gh cache list --order desc --sort size_in_bytes --limit 99999999 --json sizeInBytes | jq -r 'map(.sizeInBytes) | add' | numfmt --to=iec
5.7G
```
and I would expect this to increase over time, especially if I try other strategies to decrease CI time later.

> If you combine this with my other two suggestions, you would have an even leaner and faster CI.

Frankly these are bad suggestions. I've explained above why I can't drop Stack+Nix on every push, and I'm not going to drop old GHC versions from the Stack pipeline because that would effectively mean they are no longer supported.

At one point you mentioned you have your own fork? I think this change is a great candidate for inclusion there. If it gains traction and other people see a need for it I'm happy to revisit including it here, but I think I've made myself clear for now.

thomasjm:
All right, I'll just run this stuff in my own fork. A few more responses for what it's worth --

> Hackage heavily discourages any sort of automated release process

Automated? When did I say automated? I simply mean that you might run the Stack+Nix workflow via manual dispatch before cutting a new release. Or I don't know, configure it to run only on pushes to release branches. There are so many things you could do here besides test Stack+Nix on every push to `master`. Frankly I wouldn't have put forward #1429 at all if I had known the cache/CI time was so scarce and it would cause such problems.

> As I have pointed out before, these are additional checks that run on every commit, and additional code that someone will have to maintain over time indefinitely.

I'd just like to point out that before I started this recent set of refactors, the Nix expressions were a [messy](https://github.com/IHaskell/IHaskell/blob/1b4c2a4c3c5c6a36230aee5748d4e4569184794d/release.nix) [series](https://github.com/IHaskell/IHaskell/blob/1b4c2a4c3c5c6a36230aee5748d4e4569184794d/release-9.4.nix) of [copy-paste jobs](https://github.com/IHaskell/IHaskell/blob/1b4c2a4c3c5c6a36230aee5748d4e4569184794d/release-9.8.nix). According to `cloc` that single PR #1425 brought the repo's lines of Nix code from 511 to 385, a savings of 25%. I think I am on the side of the angels here.

If I were allowed to keep going, I would reduce the repo's Nix code still further. For example, I notice that you updated to Nixpkgs 23.11, but don't seem to have adjusted the overlays to take into account new package versions. There are likely lines now in the 9.8 overlay that can be removed; I've been meaning to look into this.

> That's only after I spent most of the day yesterday working on this. Not the most compelling argument if you ask me 😄.

Well...yes. You did good work and you improved the amount of cache headroom. Now there is room to cache more things.

> I'm not going to drop old GHC versions from the Stack pipeline because that would effectively mean they are no longer supported.

You were arguing earlier in this thread that IHaskell shouldn't be expected to support tons of GHC versions, and you've mentioned a couple times the need for a reasonable GHC deprecation policy. I agree with you. I would just suggest that such a policy isn't going to fall out the sky. You have the power to define the reasonable GHC deprecation policy. And at the current moment in time, IMO, such a policy probably would *not* support GHC 8.0, and probably *would* support 9.0.

vaibhavsagar:
> I'd just like to point out that before I started this recent set of refactors, the Nix expressions were a [messy](https://github.com/IHaskell/IHaskell/blob/1b4c2a4c3c5c6a36230aee5748d4e4569184794d/release.nix) [series](https://github.com/IHaskell/IHaskell/blob/1b4c2a4c3c5c6a36230aee5748d4e4569184794d/release-9.4.nix) of [copy-paste jobs](https://github.com/IHaskell/IHaskell/blob/1b4c2a4c3c5c6a36230aee5748d4e4569184794d/release-9.8.nix). According to cloc that single PR https://github.com/IHaskell/IHaskell/pull/1425 brought the repo's lines of Nix code from 511 to 385, a savings of 25%. I think I am on the side of the angels here.

This is unfair and a short-sighted way of looking at things. Before your refactors (which weren't purely refactors because they substantially changed how things worked), we had a couple of self-contained `release*.nix` files and a `flake.nix` file (an example of an ill-considered external contribution that I did not have the expertise to maintain over time). I could have split things up into multiple files to reduce duplication, but I always considered the newer files temporary workarounds for issues in `nixpkgs` that I would eventually be able to delete entirely so I didn't see the point (also they would no longer be self-contained).

Now actually using the Nix expressions you've written is more verbose than using the old ones was, so the line count improvement is illusory. I would argue that the new situation is in fact more complicated than the old one because I have to look at multiple files to figure out what is actually happening.

It seems like you have a particular idea of how you want the Nix expressions to work, which is different from mine, and that's completely fine, but saying you are on the side of angels is simply not true.

PR #1465, opened 2024-01-17
ihaskell-graphviz: add layout engines besides dot

sheepforce:
I've added the other graphviz layout engines besides `dot`, see https://graphviz.org/docs/layouts/


vaibhavsagar:
Thanks so much!

PR #1477, opened 2024-03-03
Dockerfile: revert to installing everything

vaibhavsagar:


vaibhavsagar:
When launched, the kernel dies due to an issue with `relayWidgetMessage`. How did this work before?

PR #1495, opened 2024-05-22
Make parseString not require Ghc monad

thomasjm:
I've been using IHaskell's [parseString](https://hackage.haskell.org/package/ihaskell-0.11.0.0/docs/IHaskell-Eval-Parser.html#v:parseString) function, and noticed that it's quite slow, taking several milliseconds just to parse a single line of code.

The reason is that it runs in the [Ghc](https://hackage.haskell.org/package/ghc-9.2.4/docs/GHC-Driver-Monad.html#t:Ghc) monad.  It's quite expensive to call [runGhc](https://hackage.haskell.org/package/ghc-9.2.4/docs/GHC.html#v:runGhc) just to parse a line of code, since it spins up a whole GHC session.

Fortunately, `parseString` doesn't actually make use of many of the `Ghc` monad's capabilities. The only functions it uses are `getSessionDynFlags` and `setSessionDynFlags`. Furthermore, it resets these flags when it's done parsing. So the only purpose for this state within the parsing code is to enable/disable extensions etc. while parsing.

Thus, we can replace the `Ghc` monad with a simple `StateT DynFlags IO`. This makes it *much* faster to do one-off parsing, with the time for me going from ~5 milliseconds to ~20 microseconds.

### DIFF
--- a/src/IHaskell/Eval/Evaluate.hs
+++ b/src/IHaskell/Eval/Evaluate.hs
@@ -21,6 +21,7 @@ module IHaskell.Eval.Evaluate (
 import           IHaskellPrelude
 
 import           Control.Concurrent (forkIO, threadDelay)
+import           Control.Monad.Trans.State (runStateT)
 import           Data.Foldable (foldMap)
 import           Prelude (head, tail, last, init)
 import qualified Data.Set as Set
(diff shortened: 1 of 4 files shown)


vaibhavsagar:
Thanks, this is great! Can I push to this branch to fix the warnings or would you like me to open a pull request to your pull request?

vaibhavsagar:
https://github.com/IHaskell/IHaskell/pull/1496/commits/bf01fe2ebe18bdd519027f5aaac9bf1790fde991 is what I had in mind, feel free to cherry-pick it or write something equivalent and I'm happy to merge this in once all tests pass.

thomasjm:
Oh sure, I can cherry-pick that.

I think we could also delete that line entirely. It looks like the intent of the parsing code is to not affect the dynamic flags at all, per [this comment](https://github.com/IHaskell/IHaskell/pull/1495/files#diff-b9fe0599c6183539e29bb1c8e6bf6f33d62d064954118ef7c73d61c08b68fa66R94).

However, the current behavior doesn't quite match the comment, as the code right after that calls `setSessionDynFlags flags`, when the actual original flags are `flags'`.  ¯\_(ツ)_/¯

vaibhavsagar:
Good catch! I'm not sure what's going on there but my suspicion is that the comment is wrong, or rather it was correct when written but a later change to the GHC API invalidated it.

PR #1524, opened 2024-10-21
Update for ipywidgets 8

abarbu:
ipywidgets 8 is the default now. The old widgets fail to run because they're out of date. They just silently produce no output in the notebook, while printing an error to the browser's console about version mismatches.

They run with these changes. I tested every widget in the example notebooks.

The basic messaging hasn't changed. Some minimum versions are different and a few fields have changed at the root of the tree. We've moved from https://github.com/jupyter-widgets/ipywidgets/blob/main/packages/schema/jupyterwidgetmodels.v7-5.md to https://github.com/jupyter-widgets/ipywidgets/blob/main/packages/schema/jupyterwidgetmodels.v8.md

The big changes have to do with tabbable, tooltips, and html descriptions being made available in most widgets. Borders was split into four, one for each direction.

NB: I did not update the example notebooks to reflect this change.


vaibhavsagar:
This looks great, thanks for doing this! Let me figure out what's happening with those failing builds (seems like a system dependency issue unrelated to this change) and I am happy to merge once that is resolved.

vaibhavsagar:
Thank you!!

PR #1525, opened 2024-10-21
Make widgets extensible

abarbu:
Jupyter widgets are extensible, but Haskell ones are not. There are alot of non-standard widgets that you might want to bind to. For example, k3d and pythreejs provide inline 3D viewers. There are all sorts of vtk based viewers for simulations. Canvas is extremely helpful. You can embed d3. You can have collapsible trees (that's in the contrib for ipywidgets).

But, we can't have any of these things because of two decisions in how ihaskell-widgets was built. The possible fields are closed by Field defined by Singletons. The possible widgets are closed by WidgetType. Using open type families we shouldn't need to impose these restrictions.

An added benefit of the changes here is that they eliminate singletons, speeding up the build massively. It used to take nearly a minute on my machine to build the Singletons module in ihaskell-widgets, the entire build takes a few seconds now. I'm not counting the large savings on dependencies as well.

Instead of data types and singletons, this just uses plain separate data types. I kept the Singletons module name so that the diff would be simpler. We define every field and widget as a separate data type without any constructors. Then use classes and open type families to relate those data types to key names and fields.

The code is simpler and more decoupled. There is hardly any promotion anymore. There is a small cost to be paid though.

The original reason for using singletons was to avoid Proxy. That's mostly true, setting fields still doesn't require Proxy. We use visible type applications instead.

But the order of the arguments had to be changed. `setField widget field val` is now `setField @Field widget val` Same with getField.

Sadly, infix operations cannot play this game. `@Field =:: val` is not valid. The solution here is to use Proxy, but to call it F for field. The code looks almost identical. You just say `F @Field =::val`.

ButtonStyle is both a field and a widget. The two share a namespace now that we don't use promotion. I named the field ButtonStyleField. Kind of ugly.

WidgetFieldPair does use a Proxy. This isn't something users should encounter much. It just made sense there.

I updated the notebooks. I did not clear them. The readme asks for that. I can if you want me to. But I'm confused as to why? Shouldn't we show off how things look?

The export list also had to change since users need more access to be able to create their own widgets. I  plan to contribute some in the future.

Note that this branch is on top of the recently submitted v8 branch since nothing works without it.


abarbu:
PS: Two other changes we could make.

One is to move to Symbol, instead of defining all of these types. That's more of a preference. Happy to follow whatever people think.

The other is to cleanly separate the core of the widget code from the particular widgets that are defined. With the new setup we could move far more code into the widgets themselves. I didn't do this to keep the PR more straightforward.

vaibhavsagar:
I'll go ahead and merge this, we can do the build fixes separately.

PR #1530, opened 2024-10-23
Fix nested buffers

abarbu:
Binary data must be passed to widgets in separate buffers instead of the main json structure.

The serialization code is incorrect for nested buffer paths like ["vertices", "data"]. This never happens in any existing widgets, they all use a non-nested path like ["values"]. We run into this problem now that we can add new widgets.

The code is supposed to strip the buffers from the json structure. The recursive case is wrong, it doesn't rebuild the original object (upd must have access to the original structure, v, not just the nested recursive case v'). Note that I only tested this in the aeson 2 case, but I did update version 1 case as well.


vaibhavsagar:
I'm not 100% following the reasoning but happy to merge this since it doesn't affect any existing widgets. Thanks!

PR #1551, opened 2025-01-21
Update `haskell/actions/setup` to `haskell-actions/setup`.

chungyc:
Other instances of `haskell/actions/setup` had already been migrated to `haskell-actions/setup`, but the one for the Docker action remained, so it is the only remaining instance of the use of the deprecated `haskell/actions/setup` action.


vaibhavsagar:
Thanks!

PR #1558, opened 2025-03-08
add an ihaskell-display module for displaying matplotlib visualizations

arossato:
Just a small module to display matplotlib visualizations through the Haskell bindings.

thanks for ihaskell, very cool!

(Add some troubles in figuring out how to install without stack: some documentation could be useful)


vaibhavsagar:
Thanks for your contribution!

> (Add some troubles in figuring out how to install without stack: some documentation could be useful)

I've chosen to avoid documenting the installation process with Cabal since it can be installed like any other package, and we already have Stack, Docker, and Nix instructions. If you would like to make a pull request to the README I would consider it.

PR #1574, opened 2025-07-29
refactor: Readability and perf optimizations to StringUtils.

mchav:
* `dropWhile isSpace` is more idiomatic.
* rstrip now avoids doing allocations from double reverse and instead only folds over the string once.


vaibhavsagar:
Thanks fam!

PR #1576, opened 2025-08-07
chore: Update ubuntu version in stack.yml workflow

mchav:
The ubuntu 20.04 is no longer supported so the stack job spins and doesn't start the github workflow.


vaibhavsagar:
Thanks! This was a long-standing issue that I was hesitant to address myself because I didn't want to be forced by our CI to drop support for GHC 8.4 and 8.6, but I guess it's time.

PR #1595, opened 2026-04-08
Add Windows support

sheaf:
This PR adds Windows support to IHaskell. I have been testing it locally on [a presentation involving a lot of plotting](https://codeberg.org/sheaf/chebyshev-talk) and it's working flawlessly.

There are several preparatory commits; in particular the commit `Avoid repeatedly calling initUnits` was crucial to improve testsuite performance as the tests took ages to run without that (see commit message for details).

The main (last) commit achieves Windows support as follows:

  - It introduces the compatibility module `IHaskell.Windows.IO` which provides the necessary handle redirection functionality that is otherwise provided by the `unix` package.
 - It updates `IHaskell.Eval.Evaluate` to use this module on Windows when communicating with the GHC session.

There's also a fair amount of faffing around with Windows vs POSIX file path separators in the testsuite, to get the testsuite green on Windows.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -1,3 +1,8 @@
+
+-- The test runner needs package environment files to supply IHaskell packages,
+-- so make sure they are always generated.
+write-ghc-environment-files: always
+
 packages: .
           ./ipython-kernel
           ./ghc-parser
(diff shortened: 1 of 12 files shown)


vaibhavsagar:
This is amazing, thanks so much!

